Issue #1, opened 2018-11-06
Zarr.jl


visr:
Hi Fabian,

Cool to see you are working on this. Didn't really go through the code yet, but will do later.

I had also made a small start, but basically the only part that was working was the metadata :)
Since I had so little and wasn't sure I'd go through with it, I didn't announce it anywhere. But here it is:
https://gitlab.com/visr/Zarr.jl

If you are up for it, we can join efforts and compare approaches.

I guess we can just call this Zarr.jl eventually, and rename Zarr.jl to ZarrPyCall or something like that?

meggart:
Hi Martijn, great to hear you are interested in this as well. Regarding the naming, I would prefer to leave the name as it is, since there are considerations on making a zarr C library zarr-developers/zarr#317. So I would like to keep the name `Zarr.jl` reserved for a wrapper around the to-be-implemented C lib. 

Regarding our approaches, it looks like we had a similar structure in mind, although my implementation seems to be a bit more advanced. It was heavily use-case driven, I just had a zarr dataset I needed to access, so I would prefer using your metadata-parsing scheme since it seems to be more principled, while mine is distributed over ZArray.jl and ZGroup.jl. 

Regarding functionality, I think the reading part is more or less working reliably (and much faster than the PyCall-based solution), and I am currently working on methods to generate Zarr arrays. 

I would also like to discuss some package design issues, in particular how to handle C and Fortran syle ordering in an efficient way. 

Do you think the package should live in JuliaGeo for discoverability or would you prefer to first develop in your or my namespace and move later either to JuliaGeo or even to zarr-developers if they are interested?

visr:
Since the spec is quite simple, I think it might be doable and preferable to have a pure julia implementation over a C binding. But this is easy to say of course :). Name is not so important right now.

I was also exploring using Dagger Chunks. Would be nice if we can take advantage of this for a higher level interface and have it integrate nicely with Dagger and JuliaDB for instance.

It's clear that you are further along. So probably best to work in this repo. I don't mind much where it lives, moving to JuliaGeo would be ok, but we can also keep it here for now, or indeed zarr-developers. I can try putting up a PR that merges in my Metadata code + tests. Since you were already using this, should we already care about breaking changes?

[...]

PR #2, opened 2018-12-12
integration of https://gitlab.com/visr/Zarr.jl

visr:
Moved the code in, but did not yet hook it up completely.
Added some tests and a minimal README.

Next planned step is to include a Metadata instance as a ZArray field, 
and integrate it further.

Note that this sacrifices FixedLengthUTF8Strings for now. It also 
removes the last type parameter of ZArray, which was the same as T 
except for FixedLengthUTF8Strings.

ref #1

If you agree with this so far, I'll continue with the rest in this branch. Didn't want to take too large a step at once.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,8 @@
+# ZarrNative.jl
+
+ZarrNative is a Julia package providing an implementation of chunked, compressed, N-dimensional arrays. [Zarr](https://zarr.readthedocs.io/en/stable/) is originally a Python package. In ZarrNative we aim to implement the [zarr spec](https://zarr.readthedocs.io/en/stable/spec/v2.html).
+
+## Links
+https://discourse.julialang.org/t/a-julia-compatible-alternative-to-zarr/11842
+https://github.com/zarr-developers/zarr/issues/284
+https://zarr.readthedocs.io/en/stable/spec/v2.html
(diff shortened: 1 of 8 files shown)


meggart:
Yes, I agree and thanks for starting this. The FixedLengthUTF8 wasn't working yet anyway so there is no loss in functionality. In the medium term we should think about string handling though or just leave it to the user to convert whatever UInt8-Array is returned by the lib so something that makes sense.

visr:
Yeah it would be nice to have something for the fixed size strings. Ideally we don't need to roll our own type for this, but there doesn't seem to be one widely used package for this.

Writing this I see that Jeff just updated https://github.com/JuliaComputing/FixedSizeStrings.jl after two years without commits though, haha. I also see https://github.com/tpapp/StrFs.jl popped up.

For LasIO we also ended up rolling our own though, https://github.com/visr/LasIO.jl/blob/wavef/src/fixedstrings.jl, but would be nice if we don't need to.

meggart:
Yes, thanks for putting this together. It is a nice place to start from and to add new functionality piece by piece. I was quite busy in the last weeks, but it looks like we will rely on a working Julia Zarr in the future, so I will put some time into this package. Maybe we can think about some kind of roadmap or todo-list so we could coordinate the next steps.

PR #3, opened 2019-01-31
Progress tutorial

meggart:
I finally started working on this package again. The original plan was to create a tutorial very similar to thi one https://zarr.readthedocs.io/en/stable/tutorial.html , just based on ZarrNative. I started with this here https://gist.github.com/meggart/63c096d454a97ce00e84caa42c615301 and just tried to fix stuff on the way that did not work, which is in this PR. 

I think it would be good to have the notebook above being converted into a doctest, so that we have some documentation and a start at testing. @visr Do you have experince with the new Documenter and documentation building under Julia 1.0? Last time I touched this was in Julia 0.5-0.6, but maybe I should simply refresh this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 Manifest.toml
+docs/build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: julia
+os:
+    - osx
+    - linux
+julia:
+    - 1.0
+    - 1.1
+notifications:
+    email: false
+jobs:
+  include:
+    - stage: "Documentation"
+      julia: 1.1
+      os: linux
+      script:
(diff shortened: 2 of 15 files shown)


visr:
Great! Indeed sounds like a good approach to recreate this tutorial, and make improvements along the way.

I worked a bit with the new Documenter for https://github.com/JuliaGeo/NetCDF.jl/pull/76. In general the [Documenter docs](https://juliadocs.github.io/Documenter.jl/stable/) are quite good, plus I followed https://discourse.julialang.org/t/psa-use-a-project-for-building-your-docs/14974.

We can make sure the documentation stays up to date with doctests, but let's also just add it to the existing unit tests, that will be easier for development, since they run with `test ZarrNative`. I took the same approach here for instance: https://github.com/JuliaGeo/GDAL.jl/blob/master/test/tutorial_vector.jl

PR #9, opened 2019-04-15
S3storage

twinGu:
Functionality to use an AWS S3 storage and read data from it.
Error in tests due to missing AWSCredentials. See https://github.com/JuliaCloud/AWSCore.jl/blob/master/src/AWSCredentials.jl
Not sure about the best way to handle this.


meggart:
It looks like the tests fail because of missing credentials. Is there any way to access the data anonymously? It looks like the Java interface has this https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/AnonymousAWSCredentials.html so do you know if there is an equivalent for Julia's AWSCore? Maybe ask in an issue?

twinGu:
I can check on Thursday if I can circumvent the authorization through AWSCore. I read your issue on there just now. I could also write some code which uses HTTP.jl directly.Am 16.04.2019 09:27 schrieb Fabian Gans <notifications@github.com>:It looks like the tests fail because of missing credentials. Is there any way to access the data anonymously? It looks like the Java interface has this https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/AnonymousAWSCredentials.html so do you know if there is an equivalent for Julia's AWSCore? Maybe ask in an issue?

twinGu:
If this https://github.com/JuliaCloud/AWSCore.jl/pull/79 gets accepted, the last commit will be working. Else maybe we could consider using the fork instead of the original repo?

meggart:
It looks like they are going to accept the PR given a unit test, so let's just wait until it gets merged and tagged  and then merge this one. Great you found that solution

twinGu:
The PR was already merged to master, but I'm not sure when a new version will be released, that will be accessible through JuliaRegistries.

meggart:
Is it possible to add the master branch as a dependency for the package? Probably not, so I would suggest to merge since the tests pass locally. Maybe you can mark the tests as [broken](https://docs.julialang.org/en/v1/stdlib/Test/index.html#Broken-Tests-1) so we will get green now and will get notified whenever tests start to pass.

twinGu:
As a workaround I added the the master branch via `Pkg.add`, which works for now. I'll open an issue, so we don't forget about this and change it back as soon as the version of AWSCore gets bumped.

PR #17, opened 2019-12-06
Aws core changes

meggart:


meggart:
Also adding some examples to the docs

PR #22, opened 2019-12-20
Add possibility to read consolidated metadata

meggart:
This would enable at least reading consolidated metadata of a Zarr storage which significantly speeds up reading remote groups with many members.


meggart:
I think this is ready to go, I refactored the PR to make ConsolidatedStore a new store type, so this should not effect existing code at all.

PR #26, opened 2020-02-03
Add compress

bagsoffelix:
the first two commits were just some testing. I am not sure about the import and line 56


src/Compressors.jl
```diff
@@ -53,12 +59,48 @@ end
 
 function zcompress(a, f::AbstractArray, ::NoCompressor)
   a2 = reinterpret(UInt8,a)
-  empty!(f)
+  empty!(f
```
meggart: This bracket should be there

PR #35, opened 2020-04-13
Support custom endpoint in S3store

manics:
`S3store` currently assumes services follow the same DNS patterns as AWS, for example `service_name.region.service_host` or `service_name.service_host`.
This means other servers such as [Minio](https://docs.min.io/docs/minio-quickstart-guide.html) can't be used. For example:
```
julia> using AWSCore

julia> using Zarr

julia> miniodev = aws_config(creds=AWSCredentials("minio", "minio123"), region="", service_host="localhost:9000", service_name="")
Dict{Symbol,Any} with 4 entries:
  :service_name => ""
  :service_host => "localhost:9000"
  :creds        => (minio, min...146138512-12-31T23:59:59)…
  :region       => ""

julia> store = Zarr.S3Store("zarr-demo", "store", aws=miniodev)
S3 Object Storage

julia> storagesize(store)
ERROR: IOError(MbedTLS error code -29184: SSL - An invalid SSL record was received during request(https://.localhost:9000/zarr-demo?list-type=2&delimiter=%2F&prefix=store%2F))

Stacktrace:
...
```

```
julia> miniodev = aws_config(creds=nothing, region="", service_host="minio-dev.openmicroscopy.org", service_name="")
Dict{Symbol,Any} with 4 entries:
  :service_name => ""
  :service_host => "minio-dev.openmicroscopy.org"
  :creds        => nothing
  :region       => ""

julia> store = Zarr.S3Store("idr", "zarr/4007817.zarr", aws=miniodev)
S3 Object Storage

julia> storagesize(store)
ERROR: DNSError: .minio-dev.openmicroscopy.org, unknown node or service (EAI_NONAME)
Stacktrace:
```

An alternative to the current `AWSSDK.S3` package is the [AWSS3 package](https://github.com/JuliaCloud/AWSS3.jl) which supports an `endpoint` parameter. I've added this as an optional backend. For example:
```
julia> miniodev = aws_config(creds=nothing, region="", endpoint="https://minio-dev.openmicroscopy.org")
Dict{Symbol,Any} with 3 entries:
  :endpoint => "https://minio-dev.openmicroscopy.org"
  :creds    => nothing
  :region   => ""

julia> store = Zarr.S3Store("idr", "zarr/4007817.zarr", aws=miniodev, backend=Zarr.awss3)
S3 Object Storage

julia> storagesize(store)
1616151347
```

Let me know what you think of this.


src/Storage/s3store.jl
```diff
@@ -76,9 +86,38 @@ end
...
+  if s.backend == awss3
+    # TODO: This doesn't list subdirectories:
+    # objects = collect(s3_list_objects(s.aws, s.bucket, prefix))
+    #
+    # https://github.com/JuliaCloud/AWSS3.jl/pull/85
+    # Dict(
+    #   "Contents" => filter(e -> haskey("Key"), objects),
+    #   "CommonPrefixes" => filter(e -> haskey("Prefix"), objects),
+    # )
+
```
manics, line 99: A problem with `AWSS3` is that it doesn't return `CommonPrefixes`. I've opened an upstream PR: https://github.com/JuliaCloud/AWSS3.jl/pull/85

test/storage.jl
```diff
@@ -92,27 +92,52 @@ import AWSCore: aws_config
...
-  @test size(arr) == (384,192,1032)
-  @test eltype(arr) == Union{Missing, Float32}
-  @test all(isapprox.(arr[1:2,1:2,2], [237.519 239.618; 237.536 239.667]))
+  for backend in [Zarr.awssdk, Zarr.awss3]
+    bucket = "cmip6"
+    store = "ScenarioMIP/DKRZ/MPI-ESM1-2-HR/ssp370/r4i1p1f1/Amon/tasmax/gn"
+    region = ""
+    if backend == Zarr.awss3
+      # AWSS3 doesn't support service_host service_name
+      aws_google = aws_config(creds=nothing, endpoint="https://storage.googleapis.com")
```
manics, line 105: Note you can't set `service_host` with `AWSS3`, you have to set the `endpoint` instead.

manics:
If you want I'm happy to run an instance of Minio inside Travis for testing. A future benefit would be that you could use it for testing S3store write support if it's added without dealing with public cloud credentials.

meggart:
Thanks a lot for the PR, though am not yet completely sold on this. The following works without modifications on master:

````julia
julia> miniodev = aws_config(creds=nothing, region="", service_host="openmicroscopy.org", service_name="minio-dev")
Dict{Symbol,Any} with 4 entries:
  :service_name => "minio-dev"
  :service_host => "openmicroscopy.org"
  :creds        => nothing
  :region       => ""

julia> store = Zarr.S3Store("idr", "zarr/4007817.zarr", aws=miniodev)
S3 Object Storage

julia> storagesize(store)
1616151347
````

However, this pattern could not solve your first use case where data is served on localhost. Back when we implemented S3Store AWSS3 was not yet flexible enough to handle these situations which is why I tried to push through this https://github.com/JuliaCloud/AWSCore.jl/pull/82 which finally got merged. 

Now that AWSS3 has the functionality, I am wondering if we could make it the only backend and remove the AWSSDK.S3 code at all. I will test more of my use cases. Note also that I personally have a project deadline next Tuesday, so it might take more than a week until I can look into this. 

Do you need this merged urgently? Also, additional tests using Minio would be great, epsecially for testing writing to S3, but let's try to got step by step.

manics:
Thanks for looking! This isn't urgent, I've only started investigating Zarr recently for bio-imaging, and since it's meant to be cross platform I thought it'd be a good excuse to teach myself Julia.

Looking at your previous PR JuliaCloud/AWSCore.jl#82 I guess one option would be to allow `service_name` to be empty, in the same way that if `region` is empty no extra `.` is added. One advantage of `AWSS3` though is it supports paging and should automatically make multiple requests if you have more than 1000 objects whereas `AWSSDK.S3` would require this library to handle that logic.

In the meantime I'll look into read-only tests for Minio.

meggart:
I think this got fixed through the new AWS interface and #55. We have Minio examples in the unit tests and docs, so I would close this one.

PR #41, opened 2020-08-10
Fixes for Julia 1.5

meggart:
This enables travis tests fir more Julia versions and fixes a deprecation on Julia 1.5


.travis.yml
```diff
@@ -6,6 +6,8 @@ os:
 julia:
     - 1.2
     - 1.3
+    - 1.4
+    - 1.5
     - nightly
```
visr, line 11: Do you prefer to test all those versions? I usually do `1.2, 1, nightly` (1 is 1.5 now)

meggart: Ok, I did not know that 1 resolves to 1.5. Yes, It probably makes sense not to test all of them. At one point we should have an appveyor test as well, in particular for 32bit systems.

meggart:
So I ran a PkgSkeleton generate on the Zarr directory to add coverage and appveyor files. Probably I still need io enable this in the repo settings, will find out.

visr:
Yeah I think you can add codecov and appveyor from their websites.

I haven't quite seen this coverage setup before, wonder why it's done like this. This PR currently still misses a way to trigger codecov, like is done here:
https://github.com/tpapp/PkgSkeleton.jl/blob/5da058b4cd40b8d51ce361399fe413169f291d69/.travis.yml#L35-L37

Though [`codecov: true`](https://docs.travis-ci.com/user/languages/julia/#coverage) seems simpler, not sure if it is 100% the same though.

meggart:
Ok, it looks like appveyor works and windows tests pass. @visr I will merge and tag if you do not complain?

visr:
Great, looks good to me!

PR #55, opened 2021-02-26
Use AWS.jl instead of AWSCore

meggart:


docs/src/s3examples.md
```diff
@@ -7,27 +7,18 @@ a s3-compatible cloud store. Here we provide examples on how to read data from c
 
 First we show how to access the zarr-demo bucket on AWS S3. We have to setup a
 AWS configuration first, for options look at the documentation of
-[AWSCore.jl](https://github.com/JuliaCloud/AWSCore.jl). If you don't have an
+[AWSCore.jl](https://github.com/JuliaCloud/AWS.jl). If you don't have an
```
visr, line 10: Link text should be updated here to AWS.jl

meggart:
Thanks for the review. Sorry, I forgot to add the WIP-label last Friday, I have added some more unit tests and a small Minio tutorial to the documentation. However, it looks like Minio.jl has julia 1.5 as a minimum requirement, which is why the tests for 1.2 fail atm. I am not sure if it is worth giving up compatibility to lower Julia versions, so I will probably remove the Minio tests again and convert the tutorial to a doctest. What do you think?

visr:
No worries, it was more of a drive by comment, not really a review. But your proposed solution seems fine to me. I wouldn't mind dropping 1.2/3/4 support either, since they are all unsupported julia releases anyway.

meggart:
Finally, after my PR got merged here https://gitlab.com/ExpandingMan/Minio.jl/-/merge_requests/1 we have working windows tests with Minio as well. For me, this would be ready to be merged. @visr feel free to take another look, otherwise I would merge in one or two days.

visr:
Great! Feel free to merge this anytime.

PR #57, opened 2021-04-16
Add datetime64 wrapper

meggart:
This adds support for date and datetime encodings to tha zarr metadata


meggart:
This is an initial draft for supporting Numpy's datetime64 types. It is a bit annoying that Julia's and Numpy's types are based on different epochs, but this seems to work. I would still have to clean up a bit and write

PR #66, opened 2021-09-30
Rewrite the storage interface

meggart:
In the past the interface for an `AbstractStore` was different to the one used in python and other Zarr implementations and led to more repetition in implementing subclasses of `AbstractStore`. To describe the difference, let's look at an example. In the previous implementation an `AbstractStore` could only point to a single Zarr Group or Array and contained the full path information in the type. When accessing a sub-array of a group, a new Store had to be created which pointed to this sub-path etc. In the new implementation a ZArray or ZGroup contains an `AbstractStore` as well as the base path inside that Storage, so that the same Storage object can be shared among different ZArrays or ZGroups in that same storage and are distinguished by their `path`s. 

This change allowed for more code sharing between different storage types, because the whole subsetting logic did not have to be re-implemented for different storages. However, it was a massive re-write and thing might initially be a bit less stable than before but I hope this will return to be more maintainable in the long run.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,11 +39,15 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
         env:
-          PYTHON: 
+          PYTHON:
       - uses: julia-actions/julia-runtest@v1
         env:
-          PYTHON: 
+          PYTHON:
       - uses: julia-actions/julia-processcoverage@v1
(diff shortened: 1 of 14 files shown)


meggart:
For now I had to limit the version of AWS.jl to 1.61 to make the tests pass, because the service interface was modified somehow. I the midterm it would be great to switch from AWS.jl to AWSS3.jl which provides convenience functions for all the operations we need to do.

PR #81, opened 2022-01-19
Fix docs URL

oxinabox:


meggart (approved):

meggart:
Thanks for fixing this, I forgot about the GH pages.

PR #86, opened 2022-03-21
Vlen utf8

meggart:
This adds support for the vlen-utf8 filter, which supports reading and writing of non-fixed length string. Unit tests and docs are still on the way.


meggart:
superseded by #116

PR #91, opened 2022-07-05
make size typestable and test for it

gdkrmr:
This makes size typestable. The step `getproperty(x, :metadata)` is still not typestable.


gdkrmr:
For documentation purposes: `Ref` is an abstract type, hence the type instability.

Issue #100, opened 2022-11-10
Reading many Zarr files (https or s3:) causes casuses computer to hand indefinitely 


alex-s-gardner:
I think this issue is related to [HTTP](https://github.com/JuliaWeb/HTTP.jl/issues/517) and could be fixed by updating the HTTP dependency

alex-s-gardner:
fixed with merge of [#101](https://github.com/JuliaIO/Zarr.jl/pull/101)

PR #101, opened 2022-11-11
Update Project.toml

alex-s-gardner:
Added and tested HTTP 1.5 as a dependency. This solves [Issue #100](https://github.com/JuliaIO/Zarr.jl/issues/100#issue-1444218904)


felixcremer:
I think that when the aim is to restrict HTTP to versions higher than 1.5 you need to delete the single 1 in the compat entry. 
Otherwise this entry allows all versions that are between 1.0.0 and lower than 2.0.0.

meggart:
Thank you! I think it would have been possible to omit the caret specifier because it is the [default anyway](https://pkgdocs.julialang.org/v1/compatibility/#Caret-specifiers) but it might be better to be explicit.

Issue #102, opened 2022-12-09
Support bitshuffle in `BloscCompressor`


nhz2:
Currently, `BloscCompressor` `shuffle` field is a `Bool`, this means compressing with bit shuffle is not possible.

https://github.com/JuliaIO/Zarr.jl/blob/c4ef0856d1fea797f1c0da6405165ec24555a6b3/src/Compressors.jl#L39-L44

 however in numcodecs,

https://numcodecs.readthedocs.io/en/stable/blosc.html shuffle is an integer:
> Either NOSHUFFLE (0), SHUFFLE (1), BITSHUFFLE (2) or AUTOSHUFFLE (-1). If AUTOSHUFFLE, bit-shuffle will be used for buffers with itemsize 1, and byte-shuffle will be used otherwise. The default is SHUFFLE.

I think the `shuffle` field type should be changed to `Int` and the logic in 
https://github.com/zarr-developers/numcodecs/blob/7d8f9762b4f0f9b5e135688b2eeb3f783f90f208/numcodecs/blosc.pyx#L264-L272 implemented in this package.

I'm happy to make a PR if you think this is a good idea.

meggart:
The reason for choosing an `Bool` here was to match the interface of [Blosc.jl](https://github.com/JuliaIO/Blosc.jl) where the shuffle argument in the docs is just a boolean. However, looking at the source code https://github.com/JuliaIO/Blosc.jl/blob/25d663c607542cbebaea45542619726fec71bb5e/src/Blosc.jl#L71-L78 indicates that integer arguments are valid as well. So yes, a PR would be most welcome, but you might have to look a bit into Blosc.jl source code to find the correct match between the numcodecs shuffle arguments and how they are passed to Blosc.jl

PR #103, opened 2022-12-15
Support bitshuffle in `BloscCompressor`

nhz2:
Implements #102 by changing the `shuffle` field of `BloscCompressor` from `Bool` to `Int`.

I think this is a non breaking change, because calling the contructor with `shuffle=true` will still result in the byte shuffle being used and  `shuffle=false` will still result in no shuffle being used, because `true` and `false` are auto converted to 1 and 0.


meggart (approved):
Looks good, thanks a lot, also for adding the python tests

PR #116, opened 2023-04-27
add support for variable-length strings

ilia-kats:
This uses the `vlen-utf8` filter, which is one of the standard filters included with Python-Zarr.

### DIFF
--- a/src/ZArray.jl
+++ b/src/ZArray.jl
@@ -11,12 +11,13 @@ const concurrent_io_tasks = Ref(50)
 getfillval(::Type{T}, t::String) where {T <: Number} = parse(T, t)
 getfillval(::Type{T}, t::Union{T,Nothing}) where {T} = t
 
-struct SenMissArray{T,N,V} <: AbstractArray{Union{T,Missing},N}
+struct SenMissArray{T,N} <: AbstractArray{Union{T,Missing},N}
   x::Array{T,N}
+  senval::T
 end
-SenMissArray(x::Array{T,N},v) where {T,N} = SenMissArray{T,N,convert(T,v)}(x)
+SenMissArray(x::Array{T,N},v) where {T,N} = SenMissArray{T,N}(x,convert(T,v))
 Base.size(x::SenMissArray) = size(x.x)
-senval(x::SenMissArray{<:Any,<:Any,V}) where V = V
+senval(x::SenMissArray) = x.senval
 function Base.getindex(x::SenMissArray,i::Int)
   v = x.x[i]
   isequal(v,senval(x)) ? missing : v
@@ -78,6 +79,7 @@ storageratio(z::ZArray{<:Vector}) = "unknown"
 
 nobytes(z::ZArray) = length(z)*sizeof(eltype(z))
 nobytes(z::ZArray{<:Vector}) = "unknown"
+nobytes(z::ZArray{<:String}) = "unknown"
 
 zinfo(z::ZArray) = zinfo(stdout,z)
 function zinfo(io::IO,z::ZArray)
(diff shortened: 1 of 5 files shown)


src/ZArray.jl
```diff
@@ -11,12 +11,13 @@ const concurrent_io_tasks = Ref(50)
 getfillval(::Type{T}, t::String) where {T <: Number} = parse(T, t)
 getfillval(::Type{T}, t::Union{T,Nothing}) where {T} = t
 
-struct SenMissArray{T,N,V} <: AbstractArray{Union{T,Missing},N}
+struct SenMissArray{T,N} <: AbstractArray{Union{T,Missing},N}
   x::Array{T,N}
+  senval::T
 end
```
meggart, line 17: I knwo this was a quite ad-hoc definition of SenMissArray, probably better to keep the sentinel value in a struct field instead of a type parameter. Now that the idea was taken up also here: https://github.com/JuliaData/SentinelArrays.jl it might be an option to just use that implementation. This should not stop this PR from being merged though.

src/ZArray.jl
```diff
@@ -361,20 +363,12 @@ function filterfromtype(::Type{<:AbstractArray{T}}) where T
   (VLenArrayFilter{T}(),)
 end
 
+filterfromtype(::Type{<:Union{<:AbstractString, Union{<:AbstractString, Missing}}}) = (VLenUTF8Filter(),)
```
meggart, line 366: Here we would default to using VLenUTF8. Do you happen to know how this is handled in the python implementation? If it the default string representation there as well, I agree with this change, just some years back I remember hitting a lot of MaxLengthStrings.

src/metadata.jl
```diff
@@ -243,4 +245,5 @@ Base.eltype(::Metadata{T}) where T = T
 fill_value_decoding(v::AbstractString, T::Type{<:Number}) = parse(T, v)
 fill_value_decoding(v::Nothing, ::Any) = v
 fill_value_decoding(v, T) = T(v)
+fill_value_decoding(v::Number, T::Type{String}) = v == 0 ? "" : T(UInt8[v])
```
meggart, line 248: Could you quickly reference an example metadata file where fill values for string arrays are encoded as numbers? I have problems to understand what is going on here.

meggart (commented):
Thanks a lot for preparing this PR. It would be really nice to have VLenUTF8 support in here. I think technically the PR would be breaking because dumping a string vector into a zarr array would result into a different representation on disk as before. So I would make sure bunp the minor version in this PR to not forget.

src/ZArray.jl
ilia-kats: I think that's because numpy by default stores string arrays as fixed-length strings. The default string representation for Python string objects is [vlen-utf8](https://github.com/zarr-developers/zarr-python/blob/main/zarr/util.py#L47), although apparently Numpy arrays with `dtype="O"` are not supported at the moment (at least it throws an exception on my system when trying to write such an array to Zarr)

src/metadata.jl
ilia-kats: One way to create something like that is to do the following:
```
import zarr
import numpy as np

test = zarr.open("test.zarr")
test2 = np.asarray(["test1", "test2"], dtype="O")
test.create_dataset("test2", shape=test2.shape, dtype=str)
test["test2"][:] = test2
```
The `.zarray` then has the following contents:
```
{
    "chunks": [
        2
    ],
    "compressor": {
        "blocksize": 0,
        "clevel": 5,
        "cname": "lz4",
        "id": "blosc",
        "shuffle": 1
    },
    "dtype": "|O",
    "fill_value": 0,
    "filters": [
        {
            "id": "vlen-utf8"
        }
    ],
    "order": "C",
    "shape": [
        2
    ],
    "zarr_format": 2
}
```

Note that this is what [AnnData is doing to store string arrays.](https://github.com/scverse/anndata/blob/main/anndata/_io/specs/methods.py#L370)

src/ZArray.jl
meggart: Ok, then let's leave it like this and make this a breaking change. I think the storage size should generally be smaller if we use variable length strings.

src/metadata.jl
meggart: Ok that would explain the first branch but I think `T(UInt8[v])` would error in all cases, because you can not index the type `UInt8`. Shall we error for non-zero numbers or just do `T(v)`?

meggart: Maybe we could add you small example in the python.jl part of the unit tests where we look at this type of interop?

ilia-kats: `UInt8[v]` creates a byte array from `v`, the assumption being that any number used as fill value is a byte sequence of a string. On my system, this runs fine:
```
julia> String(UInt8[247])
"\xf7"
```

meggart: Of course, probably I should not be as tired as yesterday when making PR reviews.

PR #117, opened 2023-04-28
add support for 0-dimensional arrays/scalars

ilia-kats:


ilia-kats:
I have to admit the spec is somewhat ambiguous with regard to zero-dimensional arrays, but the Python reference implementation has [hardcoded chunk 0](https://github.com/zarr-developers/zarr-python/blob/86a54d17e114542b34dae10e5eda1b594d40bf99/zarr/core.py#L1745), but otherwise treats zero-dimensional arrays as any other array, so I don't think the value is ever stored in the fill value.

src/Compressors.jl
```diff
@@ -2,6 +2,9 @@ import Blosc
 import CodecZlib
 import JSON
 
+_reinterpret(::Type{T}, x) where T = reinterpret(T, [x])
```
nhz2, line 5: This method seems to not be needed.

ilia-kats: it is, reinterpret doesn't work on scalars:
```julia

julia> reinterpret(UInt8, 8.)
ERROR: bitcast: argument size does not match size of target type
Stacktrace:
 [1] reinterpret(#unused#::Type{UInt8}, x::Float64)
   @ Base ./essentials.jl:438
 [2] top-level scope
   @ REPL[1]:1

julia> reinterpret(UInt8, [8.])
8-element reinterpret(UInt8, ::Vector{Float64}):
 0x00
 0x00
 0x00
 0x00
 0x00
 0x00
 0x20
 0x40
```

nhz2: But where does `_reinterpret` need to be called on a scaler? If the scalar `_reinterpret` method is deleted the tests still pass.

nhz2 (approved):
LGTM

meggart (approved):

meggart:
Thanks a lot and sorry for taking so long to review. Make sure to ping me next time if there is no reaction.

PR #118, opened 2023-06-12
Dont set connection_limit

meggart:
Currently when reading from a HTTP store a lot of warnings of this type get thrown:

````
┌ Warning: connection_limit no longer supported as a keyword argument; use `HTTP.set_default_connection_limit!(25)` before any requests are made or construct a shared pool via `POOL = HTTP.Pool(25)` and pass to each request like `pool=POOL` instead.
````
As explained in the message there are two ways to fix this. I think it would be good to not set a default at all but let the user decide what would work best for them. Is this reasonable? cc. @agoodm


agoodm:
I think this change should be fine, in my experience with the previous benchmarks the connection limit didn't make a huge difference in the results anyway.

PR #120, opened 2023-06-23
codespell: config, workflow and some typos fixed

yarikoptic:
workflow would fail whenever a new typo is introduced (or new codespell knows about some new typos).

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,4 @@
+[codespell]
+skip = .git,*.pdf,*.svg
+#
+# ignore-words-list =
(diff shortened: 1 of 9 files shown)


yarikoptic:
CI fails are on windows only and some doctest fail due to difference in `1` vs `true`.  I do not think any of those could be triggered by changes in this PR

meggart (approved):
I agree that the test failures are unrelated. Thanks a lot for the PR!

nhz2:
I think the test failures on Windows may be related to https://github.com/conda/conda/issues/9903

PR #123, opened 2023-07-14
Add read only ZipStore

nhz2:
This PR adds a new `ZipStore <: AbstractStore`  and `writezip` function.

With these, zipped Zarr data can be written to any writeable `IO` and read from any `AbstractVector{UInt8}`

This should be compatible with https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.ZipStore

Because zip files can behave differently when their size is over 4GB, this still needs to be tested with zarr-python for very large datasets.


src/Storage/Storage.jl
```diff
@@ -5,9 +5,9 @@ abstract type AbstractStore end
 
 #Define the interface
 """
-  storagesize(d::AbstractStore)
+  storagesize(d::AbstractStore, p::AbstractString)
```
sjkelly, line 8:
```suggestion
  storagesize(d::AbstractStore, [p::AbstractString])
```

This is optional so we can use square brackets to specify: https://github.com/JuliaIO/Zarr.jl/blob/cbcaeadf9d93ec174b850d90ee6db438d962f140/src/ZArray.jl#L69

src/Storage/Storage.jl
```diff
@@ -5,9 +5,9 @@ abstract type AbstractStore end
...
 
-This function shall return the size of all data files in a store.
+This function shall return the size of all data files in a store at path p.
```
sjkelly, line 10:
```suggestion
This function shall return the size of all data files in a store at path `p`.
```

src/Storage/Storage.jl
nhz2: The path is not currently optional when calling `storagesize(d::AbstractStore, p::AbstractString)`. `ZArray` is not a subtype of `AbstractStore`.

src/Storage/zipstore.jl
```diff
@@ -0,0 +1,97 @@
...
+        fullname = _make_prefix(p)*subkey
+        data = getindex(s, fullname)
+        if !isnothing(data)
+            ZipArchives.zip_writefile(w, fullname, data)
+        end
+    end
+    for subdir in subdirs(s, p)
+        _writezip(w, s, _make_prefix(p)*subdir)
+    end
+end
```
meggart, line 97: This is a very nice interface idea to produce zip stores.

test/storage.jl
```diff
@@ -145,5 +216,35 @@ end
   @test g2.attrs == Dict("groupatt"=>5)
   @test g2["a1"].attrs == Dict("arratt"=>2.5)
   @test g2["a1"][:,:] == reshape(1:200,10,20)
+  
+  # The following test doesn't pass, but maybe should?
```
meggart, line 220: Ok, I will check this later. Maybe we could make a broken test from it, let's see

meggart (approved):

meggart:
Thanks a lot and sorry for missing this for such a long time.

Issue #131, opened 2023-12-04
missing chunks to be filled with fill values (when server returns HTTP error 403)


Alexander-Barth:
When I try to load the following dataset with Zarr.jl, I get unfortunately an error:

```
using Zarr

ds = Zarr.zopen("https://s3.waw3-1.cloudferro.com/mdl-arco-time/arco/MEDSEA_MULTIYEAR_PHY_006_004/med-cmcc-cur-rean-d_202012/timeChunked.zarr")

ds["uo"][:,:,1,1]
# full error below
```

Yet, the data can be read with python zarr

```
import zarr

z = zarr.open("https://s3.waw3-1.cloudferro.com/mdl-arco-time/arco/MEDSEA_MULTIYEAR_PHY_006_004/med-cmcc-cur-rean-d_202012/timeChunked.zarr");

gz = z["uo"]
data = gz[0,0,:,:];
```

Note that all the data is filled with fill value (1e20) for this chunk. According to the [OGC spec](https://portal.ogc.org/files/100727), it seems to be ok that not all chunks are present:

```
There is no need for all chunks to be present within an array store. If a chunk is not present then
it is considered to be in an uninitialized state. An unitialized chunk MUST be treated as if it was
uniformly ﬁlled with the value of the “ﬁll_value” ﬁeld in the array metadata. If the “ﬁll_value” ﬁeld
is null then the contents of the chunk are undeﬁned.
```

Can Zarr.jl handle this case too? Are you accepting a PR for this issue?

I am using Zarr v0.9.1. 

Thank for this great package, by the way :-)

Full error from Zarr.jl:

```
ERROR: TaskFailedException
Stacktrace:
  [1] try_yieldto(undo::typeof(Base.ensure_rescheduled))
    @ Base ./task.jl:920
  [2] wait()
    @ Base ./task.jl:984
  [3] wait(c::Base.GenericCondition{ReentrantLock}; first::Bool)
    @ Base ./condition.jl:130
  [4] wait
    @ ./condition.jl:125 [inlined]
  [5] take_buffered(c::Channel{Pair{CartesianIndex{4}, Union{Nothing, Vector{UInt8}}}})
    @ Base ./channels.jl:456
  [6] take!
    @ ./channels.jl:450 [inlined]
[...]

PR #134, opened 2024-02-07
omit chunks on client HTTP error (like 404, 403,...)

Alexander-Barth:
This PR resolves issue https://github.com/JuliaIO/Zarr.jl/issues/131

### DIFF
--- a/src/Storage/http.jl
+++ b/src/Storage/http.jl
@@ -12,15 +12,24 @@ python package. In case you experience performance options, one can try to use
 """
 struct HTTPStore <: AbstractStore
     url::String
+    allowed_codes::Set{Int}
 end
+HTTPStore(url) = HTTPStore(url,Set((404,)))
 
 function Base.getindex(s::HTTPStore, k::String)
 r = HTTP.request("GET",string(s.url,"/",k),status_exception = false,socket_type_tls=OpenSSL.SSLStream)
 if r.status >= 300
-    if 400 <= r.status < 500
+    if r.status in s.allowed_codes
         nothing
     else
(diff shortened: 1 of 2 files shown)


meggart:
Sorry for missing this issue, luckily you have found out already that currently we only assume a chunk is missing when the server returns 404, which IMO should be the correct code according to e.g. this list https://en.wikipedia.org/wiki/List_of_HTTP_status_codes and so far all providers I have worked with returned this for missing chunks. 

All other codes might contain valuable information on why a request did not work, so I would also be hesitant to generally interpret these codes as missing. 

I have just pushed a different solution and hope this would work as well for you. In this PR when 403 is returned accessing a chunk you would receive an informative error message and can add the error code to the list of allowed codes that are interpreted as missing chunks. Your example would look like this:

````julia
using Revise, Zarr

ds = Zarr.zopen("https://s3.waw3-1.cloudferro.com/mdl-arco-time/arco/MEDSEA_MULTIYEAR_PHY_006_004/med-cmcc-cur-rean-d_202012/timeChunked.zarr")

ds["uo"][:,:,1,1]
#Throws informative error

Zarr.missing_chunk_return_code!(ds.storage,403)
#alternative
Zarr.missing_chunk_return_code!(ds.storage,400:499)

ds["uo"][:,:,1,1]
#No error
````

Alexander-Barth:
Perfect, yes, your solution is better. For your information, I also informed the data provider (Copernicus Marine, https://marine.copernicus.eu/) about this issue, as the HTTP error 403 (permission denied) is indeed a bit confusing.

PR #140, opened 2024-04-02
Update DiskArrays dependency to 0.4

meggart:
The whole indexing pipeline was reformulated in DiskArrays 0.4 fixing many bugs for strange vector indexing patterns. In addition this will finally make all types of getindex calls type stable. 

This is not ready to merge yet since Zarr tests revealed one upstream bug.


felixcremer:
Could this be merged now that the upstream bug is fixed and released in DiskArrays 0.4.2? 
I tested locally with DiskArrays 0.4.2 and didn't see the error anymore but have reported #144 but this also happens with DiskArrays 0.3 and seems to be independent.

PR #142, opened 2024-04-15
Change repo in docs/make.jl to fix deployment

felixcremer:
The docs have not been deployed since the move to JuliaIO because of the change in the repo url.
This should fix the deployment.


felixcremer:
The failure in nightly is unrelated and reported in #143.

PR #145, opened 2024-04-22
Factor out DateTime64 type

meggart:
I think this might be useful in other contexts as well so I put it into a small package https://github.com/meggart/DateTimes64.jl


asinghvi17:
Bump, should this be merged?

PR #149, opened 2024-06-25
Implement ZstdZarrCompressor

mkitti:
This implements ZstdZarrCompressor which wraps around CodecZstd as a package extension.

Part of the complication of using package extensions is getting a reference to new types defined in the extension. I created a mechanism by which you could specify the compressor as a string, which would then lookup the type from a dictionary.

I'm also wondering if there might be a general way to wrap TranscodingStreams codecs into Zarr compressors.

### DIFF
--- a/ext/CodecZstdExt.jl
+++ b/ext/CodecZstdExt.jl
@@ -1,0 +1,56 @@
+module CodecZstdExt
+
+using Zarr: Zarr
+using JSON: JSON
+using CodecZstd: CodecZstd
+
+"""
+    ZstdZarrCompressor(clevel::Int=0)
+    ZstdZarrCompressor(c::CodecZstd.ZstdCompressor, [d::CodecZstd.ZstdDecompressor])
+
+Zstandard compression for Zarr.jl. This is a `Zarr.Compressor` wrapper around
+`CodecZstd`. Construct with either the compression level, `clevel`, or by
+providing an instance of a `ZstdCompressor`.  `ZstdFrameCompressor` is
+recommended.
+"""
+struct ZstdZarrCompressor <: Zarr.Compressor
+    compressor::CodecZstd.ZstdCompressor
+    decompressor::CodecZstd.ZstdDecompressor
+end
+# Use default ZstdDecompressor if only compressor is provided
+function ZstdZarrCompressor(compressor::CodecZstd.ZstdCompressor)
+    return ZstdZarrCompressor(
+        compressor,
+        CodecZstd.ZstdDecompressor()
+    )
+end
+function ZstdZarrCompressor(clevel::Int)
+    return ZstdZarrCompressor(
+        CodecZstd.ZstdFrameCompressor(; level = clevel)
+    )
+end
+ZstdZarrCompressor(;clevel::Int=3) = ZstdZarrCompressor(clevel)
+
+function Zarr.getCompressor(::Type{ZstdZarrCompressor}, d::Dict)
+    return ZstdZarrCompressor(d["level"])
+end
+
+function Zarr.zuncompress(a, z::ZstdZarrCompressor, T)
+    @info "1" a
+    result = transcode(z.decompressor, a)
+    @info "2" result
+    return Zarr._reinterpret(Base.nonmissingtype(T), result)
+end
+
+function Zarr.zcompress(a, z::ZstdZarrCompressor)
+    a_uint8 = Zarr._reinterpret(UInt8,a)[:]
+    transcode(z.compressor, a_uint8)
+end
+
+JSON.lower(z::ZstdZarrCompressor) = Dict("id"=>"zstd", "level" => z.compressor.level)
+
+function __init__()
+    Zarr.compressortypes["zstd"] = ZstdZarrCompressor
+end
+
+end # module CodecZstdExt
--- a/ext/CodecZstdExt.jl
+++ b/ext/CodecZstdExt.jl
@@ -36,9 +36,7 @@ function Zarr.getCompressor(::Type{ZstdZarrCompressor}, d::Dict)
 end
 
 function Zarr.zuncompress(a, z::ZstdZarrCompressor, T)
-    @info "1" a
     result = transcode(z.decompressor, a)
-    @info "2" result
     return Zarr._reinterpret(Base.nonmissingtype(T), result)
 end
 
--- a/src/ZArray.jl
+++ b/src/ZArray.jl
@@ -333,6 +333,14 @@ function zcreate(::Type{T},storage::AbstractStore,
   attrs=Dict(),
   writeable=true,
   ) where T
+
+  if compressor isa AbstractString
+    if haskey(compressortypes, String(compressor))
+      compressor = compressortypes[compressor]()
+    else
+      throw(UnknownCompressorException(compressor))
+    end
+  end
   
   length(dims) == length(chunks) || throw(DimensionMismatch("Dims must have the same length as chunks"))
   N = length(dims)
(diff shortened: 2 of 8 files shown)


mkitti:
An alternative to the string lookup for the compressor, would be to just pass in `CodecZstd.ZstdCompressor` directly, specifically an instance created by `CodecZstd.ZstdFrameCompressor()`. Via a conversion mechanism, we could wrap that into a `Zarr.Compressor`.

ext/CodecZstdExt.jl
```diff
@@ -0,0 +1,54 @@
...
+
+Zstandard compression for Zarr.jl. This is a `Zarr.Compressor` wrapper around
+`CodecZstd`. Construct with either the compression level, `clevel`, or by
+providing an instance of a `ZstdCompressor`.  `ZstdFrameCompressor` is
+recommended.
+"""
+struct ZstdZarrCompressor <: Zarr.Compressor
+    compressor::CodecZstd.ZstdCompressor
+    decompressor::CodecZstd.ZstdDecompressor
+end
```
nhz2, line 19: This doesn't support multithreaded use IIUC. I think this should be like https://github.com/JuliaIO/Zarr.jl/blob/f436713a10a9d16ca82417176e1c8f6a3e26bc8a/src/Compressors.jl#L129-L131 where the struct only contains the parameters of the codec.

lazarusA:
bump

nhz2:
I've been working on this in https://github.com/nhz2/ChunkCodecs.jl/tree/main/ChunkCodecLibZstd

ext/CodecZstdExt.jl
meggart: Maybe I don't understand this correctly, but what would happen in a scenario where a user opens an existing array and wants to add some new data? Of course one can set a different compression level for the new chunks, but for consistency of the dataset I think it is good to write all compression parameters to the metadata struct

src/ZArray.jl
```diff
@@ -333,6 +333,14 @@ function zcreate(::Type{T},storage::AbstractStore,
   attrs=Dict(),
   writeable=true,
   ) where T
+
+  if compressor isa AbstractString
+    if haskey(compressortypes, String(compressor))
+      compressor = compressortypes[compressor]()
```
meggart, line 339: This would make it impossible to set custom compression levels for the compression algorithm. Do we need another keyword argument for `zcreate` that gets passed to the compressor constructor?

meggart (approved):
Thanks for the PR and sorry for missing it for such a long time. Probably we need to rebase and test this again. @mkitti in case you don't have the time right now I can try to rebase as well. Just let me know.

ext/CodecZstdExt.jl
nhz2: For other potential parameters what about something like:
https://github.com/nhz2/ChunkCodecs.jl/blob/799b154bd400633f0ae3bd1cf78d0cc95957f2cf/ChunkCodecLibZstd/src/encode.jl#L21-L25
```julia
struct ZstdEncodeOptions <: EncodeOptions
    compressionLevel::Cint
    checksum::Bool
    advanced_parameters::Vector{Pair{Cint, Cint}}
end
```

Where the advanced parameters are set with `ZSTD_CCtx_setParameter` after the compression level and checksum options are set.

src/ZArray.jl
mkitti: The idea here is that the simple option of just passing a string will give you default compression options. If you want to specify the compression level, you can use the compression constructor and pass the instatiated compressor instance.

mkitti:
As for how to proceed here, I do think @nhz2 has some legitimate concerns. The way I would fix the threading issues is by making a copy of the compressor and decompressed before proceeding with any operations.

The main reason for this is to keep all parameters and their validation code implemented in CodecZstd.jl rather than having to synchronize any changes here. Logic we have to encode in adapter code here will raise the maintenance burden going forward.

I'm not sure if there is an easy way to make this copy using the public Zstandard API. We might have to scan the parameters and then set them manually.

That said, I do think we should consider proceeding with this pull request mainly because it does provide baseline implementation of Zstandard for Zarr.jl, which currently does not exist.

Subsequent pull requests from either myself, Nathan, or others can revise the underlying implementation.

The most important thing to get right here is the public API this presents to the Zarr.jl user.

Another outstanding issue is whether we intend this package to implement Zarr v2 and Zarr v3, which does have some important differences with regard to how codecs are implemented, especially with regard to chains of codecs. The checksum parameters for example is now a requirement for Zarr v3 codec, but I'm not sure if it is an accepted parameter for Zarr v2.

ext/CodecZstdExt.jl
```diff
@@ -0,0 +1,54 @@
...
+    result = transcode(z.decompressor, a)
+    return Zarr._reinterpret(Base.nonmissingtype(T), result)
+end
+
+function Zarr.zcompress(a, z::ZstdZarrCompressor)
+    a_uint8 = Zarr._reinterpret(UInt8,a)[:]
+    transcode(z.compressor, a_uint8)
+end
+
+JSON.lower(z::ZstdZarrCompressor) = Dict("id"=>"zstd", "level" => z.compressor.level)
```
nhz2, line 48: This `level` property is not documented or tested in `CodecZstd` so it might be removed in a future non-breaking release. This is easy to fix by either documenting this property in `CodecZstd`, or keeping track of the compression level in the `ZstdZarrCompressor` struct.

meggart:
> Another outstanding issue is whether we intend this package to implement Zarr v2 and Zarr v3, which does have some important differences with regard to how codecs are implemented, especially with regard to chains of codecs. The checksum parameters for example is now a requirement for Zarr v3 codec, but I'm not sure if it is an accepted parameter for Zarr v2.

This is the main reason I would prefer to merge this PR soon as well. I would lreally like to start working on my v3 branch again, but would like to rebase on all outstanding PRs first, especially the ones involving codec interface changes to see what would still be compatible with v3 and how much code can be reused or would have to be adapted.

nhz2:
Yes, it makes sense to merge this now and clean up things later, though I think in that case we should mark zstd support as experimental in the docs so we can more easily change the API in the future.

bjarthur:
if i get a "ERROR: KeyError: key "zstd" not found" when opening a zarr, that's because zarr.jl doesn't support Zstd and merging this PR should resolve this error?

src/ZArray.jl
bjarthur: i get and error when trying to pass a compression constructor:

```
julia> zzeros(Int16, 16, 16, compressor=CodecZstd.ZstdCompressor(; level=6), path=tempname())
ERROR: Cannot serialize type Ptr{CodecZstd.LibZstd.ZSTD_CCtx_s}
```

mkitti:
@nhz2 Can you replace this with a ChunkCodecLibZstd.jl implementation? @bjarthur needs this as soon as possible. Otherwise, I will need to refactor this.

bjarthur:
to rephrase-- i need to read zstd-compressed slash-separated zarrs and would prefer a julia solution.

bjarthur:
this branch works for me:  https://github.com/bjarthur/Zarr.jl/tree/useme

long term it'd be nice to get this functionality into master of Zarr.jl.  Zstd is a great compressor, and hierarchical folders is much more friendly with filesystems than one big folder.

nhz2:
Yes, ChunkCodecLibZstd.jl will be registered in a few days https://github.com/JuliaRegistries/General/pull/126185
and should work like https://github.com/JuliaIO/Zarr.jl/pull/175 

The code in https://github.com/nhz2/ChunkCodecs.jl/tree/main/LibZstd and https://github.com/nhz2/ChunkCodecs.jl/tree/main/ChunkCodecCore should be good, but could have some more review.

Lastly, since blosc already depends on the zstd library, it makes sense for this to be a dependency instead of some kind of optional plugin.

mkitti:
Closed in favor of #180

PR #150, opened 2024-06-25
Attempt to implement dimension_separator

mkitti:
This is an attempt to make dimension_separator work.

Currently, Zarr.jl assumes the dimension separator is `.`.

However, N5-based variants of Zarr might have a dimension separator of `/`.

This means that a chunk might be `0/0/0` instead of `0.0.0`.

### DIFF
--- a/src/Storage/Storage.jl
+++ b/src/Storage/Storage.jl
@@ -47,17 +47,18 @@ function subkeys end
 Deletes the given key from the store.
 """
 
-citostring(i::CartesianIndex) = join(reverse((i - oneunit(i)).I), '.')
-citostring(::CartesianIndex{0}) = "0"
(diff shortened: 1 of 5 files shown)


meggart:
To me this looks like the way to go. What do you think (except missing tests) would be missing to finish this?

mkitti:
I recall being unhappy with the dimension separator lookup on every call. A better approach might be to make this a type parameter.

mkitti:
My current thinking is that we make the dimension separator a type parameter of the `AbstractStore`. Then we can have a compiled function to assemble the paths with a cached dimension separator. This will avoid a `getmetadata` function call for every single chunk, which I think might be non-neglible. Thoughts?

PR #157, opened 2024-08-22
open groups lazily

bicycle1885:
Currently, `ZGroup` opens a group immediately and recursively, which causes performance issues, especially when the group contains a large number of subarrays or subgroups. For example, in my benchmark, a single `zopen` call to open a group with 100k subarrays took 13 seconds, while a group with 1m subarrays took 266 seconds. This is a cost we want to avoid when we are only interested in a subset of those subarrays.

https://github.com/JuliaIO/Zarr.jl/blob/e3ccc9d82b20a4efeef7734d98c610ece3e0f9f3/src/ZGroup.jl#L17-L33

This PR introduces lazy group opening, meaning `zopen` now opens only the target group without recursively opening subarrays or subgroups. Subarrays and subgroups are opened on demand through `getindex`/`setindex!`. This change significantly improves performance in the scenarios mentioned above.

### DIFF
--- a/src/ZGroup.jl
+++ b/src/ZGroup.jl
@@ -1,35 +1,22 @@
 struct ZGroup{S<:AbstractStore}
     storage::S
     path::String
-    arrays::Dict{String, ZArray}
-    groups::Dict{String, ZGroup}
     attrs::Dict
     writeable::Bool
+    fill_as_missing::Bool
 end
 
 # path can also be a SubString{String}
-ZGroup(storage, path::AbstractString, arrays, groups, attrs, writeable) =
-    ZGroup(storage, String(path), arrays, groups, attrs, writeable)
+ZGroup(storage, path::AbstractString, attrs, writeable, fill_as_missing) =
+    ZGroup(storage, String(path), attrs, writeable, fill_as_missing)
 
 zname(g::ZGroup) = zname(g.path)
 
 #Open an existing ZGroup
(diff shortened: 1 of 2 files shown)


meggart:
Ok, having so many arrays in a Group is indeed problematic when eagerly opening the whole structure recursively. However, there should still be a way for users in a session to interactively explore the arrays and groups contained in a Zarr Group, as I see it there would be no way to achieve this for users given the current PR. 

Also, I am a bit concerned about performance implications for the use cases with only a few arrays in the hierarchy. With this PR, this means that for every getindex on a group the metadata would need to be parsed again. I think this should not be a big issue in many cases, but would like to check a few of my downstream applications if they are affected in some way.

bicycle1885:
Regarding the first point, I can enhance the show method to allow users to view subarrays and subgroups within the current group.

For the second point, I believe the issue won’t be problematic since the user can reuse the object returned by the initial `getindex` call. While this might require some adjustments to the user’s code, it is manageable on the user’s side. On the other hand, the slowness of `zopen` that I mentioned here cannot be easily circumvented by the user.

bicycle1885:
I added a commit to show subarrays/groups up to 10.

asinghvi17:
In the group, we agreed that we should have an option to have groups be lazy, but that should not be the default.  It creates quite some overhead. 

We can simply not populate the dictionary based on the presence of a flag.  We can even have a laziness mode which defines how to cache and whether to cache.

PR #161, opened 2024-08-26
[EXTREMELY WIP] fsspec compatible storage v2

asinghvi17:
Resurrection of the original PR #115 with a few changes for the new API.  Note that not all things work yet, and a lot of the fsspec infrastructure needs to be hooked up here.

This implementation as it currently stands has the same output as #115 including the `.zattrs` key error.  SSL errors can be prevented by passing `ssl = false` to the `FSStore` constructor.


asinghvi17:
Closing in favour of github.com/asinghvi17/FSSpec.jl

PR #165, opened 2024-09-11
Convert fill values correctly from negative to unsigned integers

asinghvi17:
Positive integers are unaffected by this change anyway.

Needs tests, probably.


meggart (commented):
Wow, how can this be intended behavior, to me it rather sounds like the files that need this are broken. Is there a publicly available dataset that has this behavior or can you share a `.zarray` file where this comes up? Just out of curiosity...

asinghvi17:
I believe I got that from https://fsspec.github.io/kerchunk/cases.html#mur-sst, the original NetCDF files have this attribute too.

meggart (approved):

lazarusA:
bump

PR #167, opened 2024-09-30
Add indents for JSON files for readbility

calvinchai:
The library writes json files minimized which is difficult to read. 
e.g.
```
{"dtype":"<u1","filters":null,"shape":[348],"order":"C","zarr_format":2,"chunks":[348],"fill_value":null,"compressor":null}
```
Added indents for readability.
```
{
    "dtype": "<u1",
    "filters": null,
    "shape": [
        348
    ],
    "order": "C",
    "zarr_format": 2,
    "chunks": [
        348
    ],
    "fill_value": null,
    "compressor": null
}
```


asinghvi17:
Should this be a choice / keyword argument instead?  Maybe it could propagate down from `write` or `zcreate`.

nhz2:
Yes, minimizing the JSON files saves disk space so adding indents should be optional.

calvinchai:
Agree. I just revised it and made it a keyword arguments

src/ZGroup.jl
```diff
@@ -128,13 +128,19 @@ end
...
     d = Dict("zarr_format"=>2)
     isemptysub(s, path) || error("Store is not empty")
     b = IOBuffer()
-    JSON.print(b,d)
+    
+    if !indent_json
+      JSON.print(b,d)
+    else
+      JSON.print(b,d,4)
+    end
```
asinghvi17, line 140:
```suggestion
    if indent_json
      JSON.print(b,d,4)
    else
      JSON.print(b,d)
    end

asinghvi17 (approved):
Looks good to me, thanks for the PR!  @meggart should probably review as well.

Test failures are from the pangeo AWS bucket that went down.

PR #168, opened 2024-10-08
add complex fill value decoding

xzackli:
Complex arrays can end up specifying a fill value of Any[0.0, 0.0], which needs to be converted into the appropriate complex number.


asinghvi17:
Thanks, could you add a test for this?

lazarusA:
bump

PR #175, opened 2025-03-02
Use `ChunkCodecLibZlib` instead of `CodecZlib`

### LINKED ISSUE #152: `ZlibCompressor` seems not to respect compression level
https://github.com/JuliaIO/Zarr.jl/blob/f436713a10a9d16ca82417176e1c8f6a3e26bc8a/src/Compressors.jl#L139-L147

It looks like the `clevel` field in the compressor is not accessed, even in `zcompress`?  Not sure if this is intentional or I am missing something.

nhz2:
Fixes #152

This also implements in-place decompression for `ZlibCompressor`.

This is using `ChunkCodecLibZlib` from https://github.com/nhz2/ChunkCodecs.jl


asinghvi17:
Sorry I missed this review request.  Broadly it looks good to me - but could we add some tests that exercise this new compressor?  Just to be sure that nothing is wrong.

asinghvi17 (approved):
Ah, it was in python.jl - did not see it there.

Looks good to me, thanks @nhz2!

PR #178, opened 2025-03-06
Add fill_value_decoding method for disambiguation

felixcremer:
This fixes the error I got in https://github.com/EOPF-Sample-Service/PMP/issues/138. 
All other methods should not be ambiguous.


felixcremer:
Yes. This still fails on 0.9.4

asinghvi17:
Retriggered CI, curious if the nightly failure is going to reappear.  Logs for that also got deleted so 🤷

PR #180, opened 2025-03-07
Add ZstdCompressor

nhz2:
Alternative to #149

This implementation supports multithreaded compression and decompression, and also supports the `checksum` option.

[ChunkCodecLibZstd](https://github.com/nhz2/ChunkCodecs.jl/tree/main/LibZstd) is being added as a direct dependency instead of a package extension, because Zarr.jl already depends on zstd through blosc.

One thing to note is that [ChunkCodecLibZstd](https://github.com/nhz2/ChunkCodecs.jl/tree/main/LibZstd) needs Julia ~~1.11~~ 1.10, and 
the ChunkCodec API is still experimental. Any suggestions for improving the API would be helpful.


bjarthur:
very much looking forward to both Zst and the multithreading in brings.  are there tests we could add to this PR to ensure it is thread safe?

nhz2:
#181 would add some basic round trip tests, which should cover all the code in this PR.

I'm not sure how to test if this is thread-safe, but in [ChunkCodecLibZstd](https://github.com/nhz2/ChunkCodecs.jl/tree/main/LibZstd), there is no global state being mutated, and the underlying C library [is supposed to be safe to use in multiple threads](https://github.com/facebook/zstd/issues/102#issuecomment-166718814).

mkitti:
Since this requires Julia 1.11 anyways, could we make this into a package extension and an optional dependency instead of a hard dependency?

The main advantage for the merge strategy here is that we do not make Zarr.jl require Julia 1.11. I would at most be more comfortable making it require Julia 1.10.

nhz2:
I'm happy to accept a PR to ChunkCodecLibZstd.jl to support Julia 1.10. Currently, the only 1.11 feature I am using is the public keyword. But is there a need to install the latest version of an in-development package on an old version of Julia?

mkitti:
If the in-development package is "Zarr.jl", then yes. Julia 1.10 is the [current long-term-support release](https://discourse.julialang.org/t/julia-v1-11-0-has-been-released-and-v1-10-is-now-lts/121064), and I would expect upcoming releases of Zarr.jl to support Julia 1.10 for some time. Making "ChunkCodecLibZstd.jl" a mandatory dependency of Zarr.jl would prevent that. I am less concerned about support for Julia versions prior to Julia 1.10.

For "ChunkCodecLibZstd.jl", dependence on Julia 1.11 is less of an issue as long as it is only an optional dependency of Zarr.jl.

[Compat.jl](https://github.com/JuliaLang/Compat.jl) could be used to address the Julia version dependency. However, I still prefer codecs as optional dependencies when possible. If a convenience package, ZarrUniverse.jl for example, is needed that loads Zarr.jl and all optional dependencies, that would not be hard to accomodate.

I will send a pull request.

mkitti:
PR for using Compat.jl for `public`: https://github.com/nhz2/ChunkCodecs.jl/pull/31
PR for making ChunkCodecLibZstd an optional dependency: https://github.com/JuliaIO/Zarr.jl/pull/183

mkitti:
I started to test the ZarrUniverse idea here:
https://github.com/mkitti/Zarr.jl/tree/mkitti-zarr-universe/lib/ZarrUniverse

```julia
using Pkg
Pkg.add(url="https://github.com/mkitti/Zarr.jl", rev="mkitti-zarr-universe", subdir="lib/ZarrUniverse")
```

or

```
] add https://github.com/mkitti/Zarr.jl#mkitti-zarr-universe:lib/ZarrUniverse
```

nhz2:
I've updated the PR. It should work with Julia 1.10 now. Also, the new `decode!` function throws a `DecodedSizeError` if the decoded size is too small or large, which cleans up the error handling.

meggart (approved):
Looks good to me, woudl be very interested to test if this works in multithreaded settings, but is defineitely not a merge-stopper since most other compressors break there as well.

mkitti:
This should be thread safe because it creates a new context for each compression and decompression call.

I understand that @bjarthur has tested these changes under a multithreaded context. It would be great to see a test for this in the test suite.

mkitti (approved):
Looks good to me.

Minor points:
1) I am wondering about the need for ChunkCodecCore to be a direct dependency here
2) Moving codecs out to package extensions can be done earlier. What Fabian discussed was perhaps making Zarr.jl the all encompassing package and then creating a core package that had optional dependencies.

PR #181, opened 2025-03-07
Fix reading complex type and add basic round trip test

nhz2:
This PR adds some basic round trip tests, which caught a bug in how complex numbers are read.

The `test/python.jl` file tests that `zarr-python` is able to read data from `Zarr.jl`. This PR adds some tests that `Zarr.jl` is able to read data from `Zarr.jl`.


nhz2:
Bump

mkitti (changes requested):
Should we make the global tuples `const` ?

PR #182, opened 2025-03-12
Add dimension separator as a type parameter

mkitti:
This replaces https://github.com/JuliaIO/Zarr.jl/pull/150 by implementing the dimension_separator as a type 
parameter.

I also intend to add Zarr specification version as a type parameter in forthcoming commits.

### DIFF
--- a/src/Storage/Storage.jl
+++ b/src/Storage/Storage.jl
@@ -3,10 +3,12 @@
 # and Dictionaries are supported
 
 """
-    abstract type AbstractStore 
+    abstract type AbstractStore{S}
 
 This the abstract supertype for all Zarr store implementations.  Currently only regular files ([`DirectoryStore`](@ref))
 and Dictionaries are supported.
+
+S is the dimension separator
 
 ## Interface
 
@@ -24,7 +26,7 @@ They may optionally implement the following methods:
 
 - [`store_read_strategy(s::AbstractStore)`](@ref store_read_strategy): return the read strategy for the given store.  See [`SequentialRead`](@ref) and [`ConcurrentRead`](@ref).
 """
-abstract type AbstractStore end
+abstract type AbstractStore{S} end
 
 #Define the interface
 """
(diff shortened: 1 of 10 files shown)


mkitti:
cc: @bjarthur @StephanPreibisch

mkitti:
@bjarthur I believe the dimension separator of `/` is now working with Zarr v2.

```julia
julia> using Zarr

julia> readdir("ts_test/zarr_v2_raw")
9-element Vector{String}:
 ".zarray"
 "0"
 "1"
 "2"
 "3"
 "4"
 "5"
 "6"
 "7"

julia> read("ts_test/zarr_v2_raw/.zarray", String) |> print
{"chunks":[8,8,8],"compressor":null,"dimension_separator":"/","dtype":"<f8","fill_value":null,"filters":null,"order":"C","shape":[64,64,64],"zarr_format":2}
julia> g = zopen("ts_test")
ZarrGroup at DirectoryStore{2, '/'}("ts_test") and path 
Variables: zarr_v2_raw 

julia> typeof(g)
ZGroup{DirectoryStore{2, '/'}}

julia> arr = g["zarr_v2_raw"]
ZArray{Float64} of size 64 x 64 x 64

julia> typeof(arr)
ZArray{Float64, 3, Zarr.NoCompressor, DirectoryStore{2, '/'}}

julia> arr[1:10,1,1]
10-element Vector{Float64}:
 0.5031057549501186
 0.30753242290354865
 0.7604873102093034
 0.3274341519944033
 0.016695504839365882
 0.4110998920786101
 0.2731260969810403
 0.5038009325254829
 0.1788864699667917
 0.8449113642679142
```

mkitti:
I have prototyped Zarr v3 on top of this and #180
https://github.com/mkitti/Zarr.jl/tree/mkitti-v3-prototype

mkitti:
Currently, I have the following storage wrapper which adds separator and zarr format version information to a storage container. Below the parameters are as follows:
* `V` is a the Zarr format version
* `SEP` is either the dimension separator (a `Char`) or a some other thing modification of the chunk key encoding
* `STORE` is the storage that is being "versioned"

```
struct VersionedStore{V,SEP,STORE <: AbstractStore} <: AbstractStore
    parent::STORE
end
```

This may be be confused for some kind of storage that keeps track of versions. I am considering renaming this to be `ForrmattedStore`. Here `CKE` replaces `SEP` since `CKE` can be any kind of chunk key encoding scheme or extension.

```
struct FormattedStore{V,CKE,STORE <: AbstractStore} <: AbstractStore
    parent::STORE
end
```

mkitti:
Resolved by merging #226

PR #183, opened 2025-04-01
Move ZstdCompressor to a package extension

mkitti:
This edits #180 to use the package extension mechanism making ChunkCodecLibZstd an optional dependency.

Note: This is PR to merge into nz/zstd-compressor, not master


mkitti:
Perhaps this should be considered as part of larger re-organization for all of the other codecs.

PR #187, opened 2025-05-08
Test against julia pre

### LINKED ISSUE #184: Test against julia prerelease
We can use pre in the julia-actions to test against the latest prerelase which would mean RC, alpha beta.

felixcremer:
This should test Zarr.jl on the current prerelease which at the moment would be 1.12 beta. 
@mkitti This is what you just requested during our meeting. 

This would close #184

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,6 +15,7 @@ jobs:
         version:
           - '1'
           - 'nightly'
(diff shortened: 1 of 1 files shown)


mkitti (approved):

nhz2:
Merging this since the new beta has fixed the code coverage error.

PR #188, opened 2025-05-08
Overload Base.UInt8 explicitly

felixcremer:
This would hopefully get rid of the following warning in julia 1.12

```julia
 ┌ Zarr
│  WARNING: Constructor for type "UInt8" was extended in `Zarr` without explicit qualification or import.
│    NOTE: Assumed "UInt8" refers to `Base.UInt8`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function UInt8 end`.
│    Hint: To silence the warning, qualify `UInt8` as `Base.UInt8` in the method signature or explicitly `import Base: UInt8`.
└  
```


mkitti:
I slipped a commit like this into #182 , but yes, doing this independently is best.

PR #194, opened 2025-06-04
Zarr v3 Prototype

mkitti:
This is a forward looking branch that builds on top of #182 to test what a Zarr v3 implementation might look like.

The initial goal is read and write valid Zarr v3 arrays that are similar conceptually to Zarr v2 arrays.


src/Storage/formattedstore.jl
```diff
@@ -0,0 +1,230 @@
...
+
+### '.' dimension separator
+
+Chunks are encoded as "c.1.2.3"
+
+### V2ChunkKeyEncoding{SEP}
+
+See Zarr version 2
+"""
+struct FormattedStore{V,SEP,STORE <: AbstractStore} <: AbstractStore
```
meggart, line 62: Why is it necessary to have the separator as a type parameter? Could this simply be a struct field?

mkitti: I earlier had tried this approach.

The main question is how much conditional evaluation do you want to do with each call to `citostring` versus constant propagation.

The separator parameter is really a "chunk key encoding" parameter. There are currently six cases:

1. Zarr v2, `'.'`
2. Zarr v2, `'/'`
3. Zarr v3, `'.'`
4. Zarr v3, `'/'`
5. Zarr v3, V2 chunk key encoding, `'.'`
6. Zarr v3, V2 chunk key encoding, `'/'`

mkitti:
Closing since [this](https://github.com/JuliaIO/Zarr.jl/pull/226/commits/9722a1a9838a2a46f3ccc2fc51fb2d94350aa553) was squashed into #226 and merged.

PR #218, opened 2025-10-20
use pipe for 1-byte eltypes

bjarthur:
fixes https://github.com/JuliaIO/Zarr.jl/issues/214#issuecomment-3420289623

PR #219, opened 2025-10-20
fix python tests on julia v1.12

bjarthur:
CondaPkg.toml now needs to be in the root dir not test/.  with this PR, https://github.com/JuliaIO/Zarr.jl/pull/216 and https://github.com/JuliaIO/Zarr.jl/pull/218 should pass.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,6 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - 'lts'
           - '1'
           - 'nightly'
           - 'pre'


mkitti:
Please add tests for to run on `"lts"`:
https://github.com/julia-actions/setup-julia?tab=readme-ov-file#examples

nhz2:
Putting CondaPkg.toml at the root means that PythonCall.jl may try to install those packages if someone is using Zarr.jl, so I think it needs to stay in the test directory.

bjarthur:
good catch @nhz2 !  i deleted the CondaPkg.toml file entirely and now using CondaPkg.add in test/python.jl to add the zarr and numpy deps.

have also updated CI.yml to test on LTS.

ready to merge?

PR #223, opened 2025-11-16
Continue v3 prototype

lazarusA:
test fix CI

TODOs

Tests:

- [x] PythonCall v3 creation
- [ ] Zarr.jl v3 creation
- [ ] compare !

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ authors = ["Fabian Gans <fgans@bgc-jena.mpg.de>"]
 version = "0.9.6"
 
 [deps]
-AWSS3 = "1c724243-ef5b-51ab-93f4-b0a88ac62a95"
 Blosc = "a74b3585-a348-5f62-a45c-50e91977d574"
 CRC32c = "8bf52ea8-c179-5cab-976a-9e18b702a9bc"
 ChunkCodecCore = "0b6fb165-00bc-4d37-ab8b-79f91016dbe1"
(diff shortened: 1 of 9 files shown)


mkitti (changes requested):
I need to check the shadd index. I think it is transposed from what I would expect it to be.

src/Codecs/V3/V3.jl
```diff
@@ -0,0 +1,531 @@
...
+end
+
+"""
+    set_chunk_slice!(idx::ShardIndex, chunk_coords::NTuple{N,Int}, offset::Int, nbytes::Int)
+
+Set the byte range for a chunk in the index.
+"""
+function set_chunk_slice!(idx::ShardIndex, chunk_coords::NTuple{N,Int}, offset::Int, nbytes::Int) where N
+    idx.chunks[chunk_coords...] = ChunkShardInfo(UInt64(offset), UInt64(nbytes))
+end
```
mkitti: Perhaps we should defer the casting to `UInt64` to the `ChunkShardInfo` constructor.

Also, I think we could relax the arguments to `Integer` rather than `Int`. Also, I am definitely sure `Int` should not depend on the system word size (`Int64` or `Int32`).

src/ZArray.jl
```diff
@@ -321,48 +322,70 @@ Creates a new empty zarr array with element type `T` and array dimensions `dims`
...
-  chunks,
-  typestr(T),
-  compressor,
-  fill_value,
-  'C',
-  filters,
+  
+  # Create a dummy array to use with Metadata constructor
+  # This allows us to leverage the multiple dispatch in Metadata constructors
+  dummy_array = Array{T,N}(undef, dims...)
```
meggart: Please don't do this. We might be creating a TB-sized array here...

PR #226, opened 2025-11-28
Continue on v3 PR

meggart:
This is my branch based on the changes by @mkitti and @lazarusA. The main change I did is that I removed the concept of the FormattedStore again for now and rather put the task of transforming chunk Cartesian Indices into string into the hand of the calling function. 

I think when I started the package it was not a good decision to put some of the Cartesian Index logic into the Storage part of the package because this is rather an encoding step and independent of the storage backend. 

This is not yet super cleaned up, but I think the goal should be to have storage backend function like `getindex`, `setindex!` etc only operate on string keys, which is why I renamed the CartesianIndex-related functions to indicate that additional information is needed to parse them. 

Also for helper functions like `writeattrs` which definitely I would prefer we explicitly pass the zarr version information instead of wrapping stores into v2 and v3. @mkitti could you live with such a solution? If not we could as well return to using some `FormattedStore` type and attach the way chunks are encoded to the storage type, but somehow I would prefer if `AbstractStore` remained a very low-level abstraction for dealing with basic IO operations, independent of Zarr versions and encoding options.

### DIFF
--- a/src/metadata.jl
+++ b/src/metadata.jl
@@ -39,11 +39,6 @@ typestr(::Type{<:Array}) = "|O"
 typestr(t::Type{<:DateTime64}) = pydatetime_string(t)
 typestr(::Type{<:AbstractString}) = "|O"
 
-function typestr(t::Type, big_endian::Bool)
-    s = typestr(t)
-    big_endian ? replace(s, '<' => '>') : s
-end
-
 const typestr_regex = r"^([<|>])([tbiufcmMOSUV])(\d*)(\[\w+\])?$"
 const typemap = Dict{Tuple{Char, Int}, DataType}(
     ('b', 1) => Bool,
@@ -59,19 +54,6 @@ foreach([Float16,Float32,Float64,Int8,Int16,Int32,Int64,Int128,
   UInt8,UInt16,UInt32,UInt64,UInt128,
   Complex{Float16},Complex{Float32},Complex{Float64}]) do t
     typemap[(typecharf(t),sizemapf(t))] = t
-end
-
-function is_big_endian(s::AbstractString)
-    m = match(typestr_regex, s)
-    if m === nothing
-        throw(ArgumentError("$s is not a valid numpy typestr"))
-    end
-
-    byteorder, typecode, typesize, typespec = m.captures
-    if byteorder == ">"
-        return true
-    end
-    return false
 end
 
 function typestr(s::AbstractString, filterlist=nothing)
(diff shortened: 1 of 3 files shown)


mkitti:
One application is transitioning a V2 array to a V3 array. You can do this by just adding a zarr.json to an existing V2 array without moving the chunks by using a V2 chunk-key-encoding in the zarr json.

There are also other possible chunk-key-encoding extensions. For example, I have proposed a suffix chunk-key-encoding zarr-extension where chunks may have a file extension:
https://github.com/zarr-developers/zarr-extensions/pull/28

Chunks with file extensions are interesting because the chunks themselves could be PNG files or even (Geo)TIFFs.

mkitti (changes requested):
We need to abstract `ChunkKeyEncoding` out because V3 arrays can use a V2 chunk key encoding, also there are chunk-key-encoding extensions.

src/metadata.jl
```diff
@@ -86,75 +86,171 @@ function typestr(s::AbstractString, filterlist=nothing)
...
+zarr_format(::MetadataV2) = ZarrFormat(Val(2))
+
+"""Metadata for Zarr version 3 arrays"""
+struct MetadataV3{T,N,C,F} <: AbstractMetadata{T,N,C,F}
+    zarr_format::Int
+    node_type::String
+    shape::Base.RefValue{NTuple{N, Int}}
+    chunks::NTuple{N, Int}
+    dtype::String  # data_type in v3
+    compressor::C
```
mkitti: If we are going to implement V3 metadata properly, we could probably remove this and filters and replace with `Codecs`.

src/metadata.jl
```diff
@@ -86,75 +86,171 @@ function typestr(s::AbstractString, filterlist=nothing)
...
 end
+zarr_format(::MetadataV2) = ZarrFormat(Val(2))
+
+"""Metadata for Zarr version 3 arrays"""
+struct MetadataV3{T,N,C,F} <: AbstractMetadata{T,N,C,F}
+    zarr_format::Int
+    node_type::String
+    shape::Base.RefValue{NTuple{N, Int}}
+    chunks::NTuple{N, Int}
+    dtype::String  # data_type in v3
```
mkitti: Might as well call this `data_type`.

mkitti (changes requested):
I think I see better what are you trying to do with the chunk-key-encoding, but I think this is going to be problematic in the long run if we do not make this extensible.

Regarding the V3 Metadata structure, I think this is going to require much more extensive changes to get correct. It should be quite different from the V2 Metadata model since codecs can take on a tree-like structure. The sharding codec for example can have multiple nested levels. There is really no distinction between filters and compressors.

mkitti:
Hey guys, any thoughts? Should try to meet Thursday?

Issue #63, opened 2021-08-02
Big-endian data


tjdiamandis:
Hello,

How difficult is it to add Big-endian functionality for `zopen`? Happy to work on this if some guidance is provided.

Thank you!

https://github.com/meggart/Zarr.jl/blob/3aaf6d0555b0a69a585da5c69334eedef6ee2b51/src/metadata.jl#L90

meggart:
Thanks a lot for offering help here. 

The first step to support big-endian data would be to first remove the exception so that arrrays can be opened. Then I would try to hook into the `readchunk!` and `writechunk!` functions (defined in ZArray.jl), because every array access passes through these functions. So it should be sufficient to add a check in `readchunk!` that tests for endianness in the array's metadata and applies `ntoh` to the output array `a` in-place if necessary. The same could be done for `writechunk!` as well. 

Let me know if that already helps or if you need more details.

tjdiamandis:
Makes sense. Thank you very much! I'll take a look soon.

lazarusA:
closed by #230

PR #230, opened 2026-01-25
Add read support for big endian

lupemba:
Add support for zarr datasets with big-endian data https://github.com/JuliaIO/Zarr.jl/issues/63

The PR adds 
- Read support for data with big-endian
- Write support for data with big-endian
- Test case with big-endian data


lupemba:
@felixcremer
I got Claude to help me add a tiny test dataset (even tiny compared to the code). I needed python Zarr version 3 to generate it so I made a separate script to do it. The python.jl test file only works with python Zarr version 2.

lupemba:
This PR is made with the assistance of AI (Claude code) but it is reviewed, modified and tested by me.

The PR adds write and read support for big-endian variables. I am currently using the read part on a large dataset and it works flawlessly for me. The write part I added for completeness. I have tested the write part but I am not using it regularly.

I have made a small test against a reference created with Python Zarr  v3. I get an identical output with Zarr.jl except for 
`"dimension_separator"` which is not set in Zarr.jl and `"fill_value"` which is set to `nothing` in Zarr.jl. I believe that this are due to different defaults and are unrelated to this PR. 

Script to make python reference
```python
"""Generate a small test zarr file with big-endian data (zarr v2 format)."""

import numpy as np
import zarr
import shutil
from pathlib import Path

# Output directory
output_dir = Path(__file__).parent / "big_endian_test.zarr"

# Clean up if exists
if output_dir.exists():
    shutil.rmtree(output_dir)

# Create a group with a named array with big-endian int16 dtype (zarr v2 format)
data = np.arange(10, dtype=">i2")
root = zarr.open_group(str(output_dir), mode="w", zarr_format=2)
z_write = root.create_array("big_endian_var", shape=10, dtype=">i2", chunks=10)
z_write[:] = data

```

lupemba:
ping @felixcremer @meggart

PR #252, opened 2026-03-23
[WIP] Extract ZarrCore as a separate library

### LINKED ISSUE #186: Refactor Zarr.jl into ZarrCore (minimal pure Julia implementation) and Zarr.jl could load all the extra packages.
The ZarrCore package would be for power users that know exactly which package extensions they need for their codecs and Zarr.jl would stay for normal users. 
The ZarrCore package should live as a subpackage in the Zarr.jl repository.

asinghvi17:
Fixes #186.

The idea of this pr is to extract a core library ZarrCore.jl so that it can be (a) used in Zarr and (b) potentially depended on by Zarrs.jl also.  This way,
- we can have a higher degree of interoperability and swappability between the two libraries - so Zarr.jl could use a store defined in Zarrs.jl, for example, and they would subtype the same supertype.
- for people who require 0 binary dependencies, we have a pure Julia version of zarr

Below follows a more detailed changelog.  I'm still testing this and not quite there with the API yet.  But it seems to work so far.

  - Extract a minimal ZarrCore.jl package at lib/ZarrCore/ containing all core types, registries, and pure-Julia implementations sufficient to read/write uncompressed Zarr v2 and  
  v3 arrays
  - Rewrite Zarr.jl as a thin wrapper that depends on ZarrCore, registering Blosc/Zlib/Zstd compressors, V3 compression codecs, and network/archive stores                          
  - Replace hardcoded codec parsing in metadata3.jl with extensible v3_codec_parsers registry and codec_to_dict dispatch


# Detailed description

 ## What's in ZarrCore

  - Types: ZarrFormat, AbstractCodecPipeline, MetadataV2, MetadataV3, V2Pipeline, V3Pipeline, ZArray, ZGroup                                                                        
  - Registries: compressortypes, v3_codec_parsers, codec_to_dict, filterdict, storageregexlist, chunk_key_encoding_parsers
  - Compressor: NoCompressor only (default_compressor() returns it)                                                                                                                 
  - V3 Codecs: BytesCodec, TransposeCodec                                                                                                                                           
  - Filters: All pure-Julia (Fletcher32, Shuffle, Delta, Quantize, FixedScaleOffset, VLen)                                                                                          
  - Stores: DirectoryStore, DictStore, ConsolidatedStore                                                                                                                            
  - API: zcreate, zopen, zzeros, zgroup, consolidate_metadata                                                                                                                       
                                                                  
 ## What stays in Zarr.jl                                                                                                                                                             
                                                                  
  - Blosc/Zlib/Zstd compressors (override default_compressor() → BloscCompressor())                                                                                                 
  - V3 compression codecs: GzipV3Codec, BloscV3Codec, ZstdV3Codec, CRC32cV3Codec, ShardingCodec
  - Network/archive stores: HTTPStore, GCStore, S3Store, ZipStore                                                                                                                   
  - AWSS3 extension (unchanged)                                   

                                                                                                                                                   
 ## Key design decisions
                                                                                                                                                                                    
  - __init__ registration pattern: All Dict/Array mutations (compressor types, codec parsers, storage regexes) happen in Zarr.__init__() since top-level mutations don't persist    
  through Julia precompilation
  - DEFAULT_COMPRESSOR_FACTORY Ref: Allows Zarr.jl to override ZarrCore's default compressor at runtime                                                                             
  - Registry-based V3 codec parsing: v3_codec_parsers Dict replaces the hardcoded if/elseif chain in metadata3.jl, making codecs extensible without modifying core code             
  - codec_to_dict dispatch: Replaces hardcoded isa checks in lower3 serialization
  - Backward-compatible module paths: Zarr.


asinghvi17:
I'm going to close this and do a new PR, since we have had some substantive internal changes.

mkitti:
Maybe we should try a different approach here. Currently, this branch is just going to be constantly out of date as merge other pull requests.

How about we just start with an empty package as a new dependency and register that. Then migrate one thing at a time?

asinghvi17:
Yeah that would definitely help.  I can scaffold that out today

mkitti:
I think we should revisit this as an objective in Germany.

asinghvi17:
That sounds good.  I don't imagine I will have time to work on this until Juliacon begins.

Issue #247, opened 2026-03-16
zarr_format in groups


lazarusA:
while creating the groups example, 

```Julia
using Zarr

store = Zarr.DirectoryStore("experiment.zarr")
g = zgroup(store, "", Zarr.ZarrFormat(3))

zcreate(Float64, g, "temperature", 100, 100; chunks=(50, 50), fill_value=0.0)
zcreate(Float64, g, "precipitation", 100, 100; chunks=(50, 50), fill_value=0.0)
g
```

I noticed that if we don't explicitly say `zarr_format=3` in `zcreate` then it defaults to `.zarray` and `.zattrs`. I think the version should be inherit from `g`.

_Originally posted by @lazarusA in https://github.com/JuliaIO/Zarr.jl/issues/246#issuecomment-4065234245_

PR #253, opened 2026-04-02
Add support for S3Path and remove deprecated global_aws_config()

lupemba:
- Add `zopen` method for `AWS3.S3Path`. 
- Replace `AWSS3.global_aws_config` because of deprecation warnings.

### DIFF
--- a/docs/src/tutorials/s3examples.md
+++ b/docs/src/tutorials/s3examples.md
@@ -12,13 +12,15 @@ account, you can access the dataset without credentials as follows:
 
 ````@example aws
 using Zarr, AWSS3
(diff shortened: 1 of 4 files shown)


lazarusA:
Thanks! LGTM.

lazarusA (approved):

lupemba:
@lazarusA 
Thanks for merging. It have been some time since the last Zarr.jl release. Would it make sense to tag a new version?

lazarusA:
I was hopping for #241 to be finished before that, and maybe fix #247, so that zarr version 3 is mostly complete. At the moment I don't have a lot of bandwidth to dive into any of those, unfortunately.

PR #258, opened 2026-04-10
added zarr_format to ZGroup

### LINKED ISSUE #247: zarr_format in groups

lazarusA:
closes #247

### DIFF
--- a/src/ZGroup.jl
+++ b/src/ZGroup.jl
@@ -5,11 +5,12 @@ struct ZGroup{S<:AbstractStore}
     groups::Dict{String, ZGroup}
     attrs::Dict
     writeable::Bool
+    zarr_format::ZarrFormat
 end
 
 # path can also be a SubString{String}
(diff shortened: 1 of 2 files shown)


mkitti (approved):
We do need this

Issue #254, opened 2026-04-02
Support for GeoZarr conventions


pvanlaake:
GeoZarr is built on top of Zarr to provide semantics to the geospatial dimensions of zarr arrays. The mechanics of such support is through the use of so-called conventions: community-based conventions to encode a certain domain in a standards-based manner into GeoZarr. The process and the initial proposed conventions can be found on the [GeoZarr homepage](https://geozarr.org).

There is an active community process on-going to further develop GeoZarr along these lines, with an eye to registering GeoZarr as an OGC standard for geospatial data. Currently, there are efforts to implement GeoZarr conventions in Python and R and we would love to see the Julia community engage as well. Apart from implementing the proposed conventions, you are also very welcome to join the discussions on taking GeoZarr further.

Looking forward to seeing you there and ready to answer any questions you might have in this thread.

lazarusA:
@felixcremer should we create a `ZarrConventions.jl` package and defined them all there? https://geozarr.org/conventions.html ? and then create an `ext` here for them? Thoughts...

pvanlaake:
If anyone is interested, there is another GeoZarr SWG meeting today Wednesday at 16:00 UTC. Details can be found on the GeoZarr [home page](https://geozarr.org).

felixcremer:
I will most likely not be able to make it to the meeting today, because it is dinner time but I started implementing reading of GeoZarr data in https://github.com/JuliaDataCubes/YAXArrayBase.jl/pull/56.

asinghvi17:
This should be outside the Zarr.jl repo.  But we should do it at some stage.

PR #261, opened 2026-04-21
Check whether must_understand is set to false

felixcremer:
If I am understanding the zarr spec correctly metadata entries that contain a must_understand key set to false are discouraged but allowed. https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#extensions-section

This sets a check whether must_understand is set to false if a key is not in the default keys.
I came across this while working on #254  and on the example data there is a consolidated_metadata field in the attributes which contains a must_understand set to false.

### DIFF
--- a/src/metadata3.jl
+++ b/src/metadata3.jl
@@ -173,7 +173,11 @@ function Metadata3(d::AbstractDict, fill_as_missing)
         # Optionally they can have attributes
         for key in keys(d)
             if key ∉ ("zarr_format", "node_type", "attributes")
-                throw(ArgumentError("Zarr v3 group metadata cannot have a key called $key"))
+            if d[key]["must_understand"] == false
+                @warn "Zarr v3 group metadata has an unrecognized key called $key with must_understand=false; ignoring"
+            else
+                throw(ArgumentError("Zarr v3 group metadata has an unrecognized key called $key with must_understand=true"))
+            end
             end
         end
 
@@ -201,7 +205,12 @@ function Metadata3(d::AbstractDict, fill_as_missing)
     check_keys(d, mandatory_keys)
     for key in keys(d)
         if key ∉ mandatory_keys && key ∉ optional_keys
-            throw(ArgumentError("Zarr v3 metadata cannot have a key called $key"))
+            if d[key]["must_understand"] === false
+                @warn "Zarr v3 array metadata has an unrecognized key called $key with must_understand=false; ignoring"
+            else                
+                throw(ArgumentError("Zarr v3 array metadata has an unrecognized key called $key with must_understand=true"))
+            end
+            #throw(ArgumentError("Zarr v3 metadata cannot have a key called $key"))
         end
     end
 
--- a/src/metadata3.jl
+++ b/src/metadata3.jl
@@ -153,7 +153,8 @@ function get_order(md::MetadataV3)
 end
 get_order(md::MetadataV2) = md.order
 
-
+must_understand(x) = true # The default of must_understand is true
+must_understand(x::Dict) = get(x, "must_understand", true)
 
 function Metadata3(d::AbstractDict, fill_as_missing)
     check_keys(d, ("zarr_format", "node_type"))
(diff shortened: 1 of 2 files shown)


src/metadata3.jl
```diff
@@ -173,7 +174,11 @@ function Metadata3(d::AbstractDict, fill_as_missing)
         # Optionally they can have attributes
         for key in keys(d)
             if key ∉ ("zarr_format", "node_type", "attributes")
```
lazarusA, line 176: probably is best to include this as another key here.

src/metadata3.jl
```diff
@@ -201,7 +206,12 @@ function Metadata3(d::AbstractDict, fill_as_missing)
     check_keys(d, mandatory_keys)
     for key in keys(d)
         if key ∉ mandatory_keys && key ∉ optional_keys
```
lazarusA, line 208: and add it here as a optional `key`.

src/metadata3.jl
felixcremer: No, that is not how this is supposed to work.

src/metadata3.jl
```diff
@@ -153,7 +153,8 @@ function get_order(md::MetadataV3)
 end
 get_order(md::MetadataV2) = md.order
 
-
+must_understand(x) = true # The default of must_understand is true
+must_understand(x::Dict) = get(x, "must_understand", true)
```
mkitti, line 157: This could be AbstractDict. We may need to use OrderedDict in the future.

src/metadata3.jl
```diff
@@ -173,7 +174,11 @@ function Metadata3(d::AbstractDict, fill_as_missing)
         # Optionally they can have attributes
         for key in keys(d)
             if key ∉ ("zarr_format", "node_type", "attributes")
-                throw(ArgumentError("Zarr v3 group metadata cannot have a key called $key"))
+            if !must_understand(d[key])
```
mkitti, line 177: This suggests the existence of an additional top level key in zarr.json, but that is technically not valid. All user level metadata should be under the attributes key or at defined extension points.

That said, I have seen a consolidated metadata key at the top level. I also think that should be under attributes.

src/metadata3.jl
```diff
@@ -201,7 +206,12 @@ function Metadata3(d::AbstractDict, fill_as_missing)
     check_keys(d, mandatory_keys)
     for key in keys(d)
         if key ∉ mandatory_keys && key ∉ optional_keys
-            throw(ArgumentError("Zarr v3 metadata cannot have a key called $key"))
+            if d[key]["must_understand"] === false
```
mkitti, line 209: Shouldn't you use the `must_understand` function here? What if the key does not exist?

test/v3_codecs.jl
```diff
@@ -394,6 +394,9 @@ end
 
     # Extra key in group metadata
     @test_throws ArgumentError Zarr.Metadata("""{"zarr_format":3,"node_type":"group","bad_key":1}""", false)
+    
+    #Extra dict that is ignored
+    @test_warn "ignoring" Zarr.Metadata("""{"zarr_format":3,"node_type":"group","ignore_dict":{"must_understand":false}}""", false) isa Zarr.Metadata
```
mkitti, line 399: An extra top level key should be an error.

mkitti (changes requested):

d-v-b:
the spec is currently a bit in flux on this point. see https://github.com/zarr-developers/zarr-specs/issues/371. You can choose between strictly adhering to the current language, which forbids an additional key with the `{"must_understand": False, ...}` key: value pair, but this makes consolidated metadata as implemented by zarr-python invalid. Or you can allow a wider type than what the spec currently allows (namely, the type that _used_ to be allowed).

IMO this is a high priority for resolution at the spec level.

test/v3_codecs.jl
felixcremer: My understanding is, that a toplevel key is discouraged but possible, here the text from the spec:

Use of must_understand=False to add top-level keys is discouraged in favor of the explicit use of [Extension points](https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#extension-points).

mkitti:
At the moment, I have not seen top-level metadata other than consolidated metadata. At the top level, I want to make a special case for consolidated metadata.

We can implement must_understand below the top level.

d-v-b:
> At the moment, I have not seen top-level metadata other than consolidated metadata. At the top level, I want to make a special case for consolidated metadata.

I'm also unaware of any top-level extra keys beyond consolidated metadata.

PR #269, opened 2026-05-07
do not allocate in-memory array just for metadata construction

bjarthur:
as it stands now it is impossible to create a Zarr array that won't fit in memory.  because a temporary array is create solely to initialize the metadata.  this PR fixes that by defining an empty array type that has the same eltype and returns the fake size when queried.

the alternative would be to define more Metadata methods, but that's a much more invasive change.

@mkitti

### DIFF
--- a/src/ZArray.jl
+++ b/src/ZArray.jl
@@ -342,6 +342,8 @@ end
 struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
     sz::NTuple{N,Int}
 end
+ShapeOnlyArray{T,N}(sz::NTuple{N,<:Integer}) where {T,N} =
+    ShapeOnlyArray{T,N}(ntuple(i -> Int(sz[i]), N))
 Base.size(a::ShapeOnlyArray) = a.sz
 Base.getindex(::ShapeOnlyArray, ::Vararg{Int}) =
     error("ShapeOnlyArray carries no data")
--- a/src/ZArray.jl
+++ b/src/ZArray.jl
@@ -343,7 +343,7 @@ struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
     sz::Dims{N}
 end
 Base.size(a::ShapeOnlyArray) = a.sz
-Base.getindex(::ShapeOnlyArray, ::Vararg{Int}) =
+Base.getindex(::ShapeOnlyArray, ::Vararg{Any}) =
     error("ShapeOnlyArray carries no data")
 
 function zcreate(::Type{T},storage::AbstractStore,
--- a/src/ZArray.jl
+++ b/src/ZArray.jl
@@ -342,8 +342,6 @@ end
 struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
     sz::Dims{N}
 end
-ShapeOnlyArray{T,N}(sz::NTuple{N,<:Integer}) where {T,N} =
-    ShapeOnlyArray{T,N}(ntuple(i -> Int(sz[i]), N))
 Base.size(a::ShapeOnlyArray) = a.sz
 Base.getindex(::ShapeOnlyArray, ::Vararg{Int}) =
     error("ShapeOnlyArray carries no data")
--- a/src/ZArray.jl
+++ b/src/ZArray.jl
@@ -340,7 +340,7 @@ function zcreate(::Type{T}, dims::Integer...;
 end
 
 struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
-    sz::NTuple{N,Int}
+    sz::Dims{N}
 end
 ShapeOnlyArray{T,N}(sz::NTuple{N,<:Integer}) where {T,N} =
     ShapeOnlyArray{T,N}(ntuple(i -> Int(sz[i]), N))
--- a/src/ZArray.jl
+++ b/src/ZArray.jl
@@ -339,6 +339,13 @@ function zcreate(::Type{T}, dims::Integer...;
   zcreate(T, store, dims...; zarr_format, dimension_separator, kwargs...)
 end
 
+struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
+    sz::NTuple{N,Int}
+end
+Base.size(a::ShapeOnlyArray) = a.sz
+Base.getindex(::ShapeOnlyArray, ::Vararg{Int}) =
+    error("ShapeOnlyArray carries no data")
+
 function zcreate(::Type{T},storage::AbstractStore,
   dims...;
   path = "",
@@ -370,7 +377,7 @@ function zcreate(::Type{T},storage::AbstractStore,
   
   # Create a dummy array to use with Metadata constructor
   # This allows us to leverage the multiple dispatch in Metadata constructors
-  dummy_array = Array{T,N}(undef, dims...)
+  dummy_array = ShapeOnlyArray{T,N}(dims)
   metadata = Metadata(dummy_array, chunks, v;
       compressor=compressor,
       fill_value=fill_value,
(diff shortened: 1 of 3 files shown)


src/ZArray.jl
```diff
@@ -370,7 +377,7 @@ function zcreate(::Type{T},storage::AbstractStore,
   
   # Create a dummy array to use with Metadata constructor
   # This allows us to leverage the multiple dispatch in Metadata constructors
-  dummy_array = Array{T,N}(undef, dims...)
```
lazarusA, line 373: Ohhh 😧, this line change got in?? I didn't catch that. No good. Thanks for the fix. I wonder if the proposed change is how it was done before in 0.9.

bjarthur: your commit that changed this @lazarusA is [here](https://github.com/JuliaIO/Zarr.jl/commit/db8a08c01a4c985bd0ac441c23ae00760be3ae57).  you seemed then to want to use dispatch.  you like this PR or just want to revert that commit?

lazarusA: I like your approach. But, let's also see if @mkitti is ok with it.

lazarusA (approved):

src/ZArray.jl
mkitti: This works. I think it would be slightly more straightforward to pass the eltype and size directly, but I wonder if someone has overriden this method elsewhere for a DiskArray or something.

src/ZArray.jl
```diff
@@ -339,6 +339,13 @@ function zcreate(::Type{T}, dims::Integer...;
   zcreate(T, store, dims...; zarr_format, dimension_separator, kwargs...)
 end
 
+struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
+    sz::NTuple{N,Int}
```
mkitti:
```suggestion
    sz::Dims{N}
```

mkitti (changes requested):
Small change

bjarthur:
why is there no "re-request review" next to copilot?

mkitti:
You might need Copilot Pro access. Consider applying for free access via https://docs.github.com/en/copilot/how-tos/copilot-on-github/set-up-copilot/enable-copilot/set-up-for-teachers-and-os-maintainers#accessing-copilot-pro-for-free

src/ZArray.jl
```diff
@@ -339,6 +339,13 @@ function zcreate(::Type{T}, dims::Integer...;
   zcreate(T, store, dims...; zarr_format, dimension_separator, kwargs...)
 end
 
+struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
+    sz::Dims{N}
+end
+Base.size(a::ShapeOnlyArray) = a.sz
+Base.getindex(::ShapeOnlyArray, ::Vararg{Any}) =
+    error("ShapeOnlyArray carries no data")
```
mkitti, line 347: code coverage is complaining that you have not tested this.

mkitti:
@copilot please review again. Suggest how to test the new getindex method.

test/runtests.jl
```diff
@@ -256,6 +256,17 @@ end
   @test_throws ArgumentError resize!(a,(-1,2))
 end
```
mkitti, line 278:
```suggestion
@testset "ShapeOnlyArray" begin
    A = ShapeOnlyArray{Float64,3}((3,4,5))
    @test size(A) == (3,4,5)
    @test eltype(A) == Float64
    @test_throws ErrorException A[1]
    @test_throws ErrorException A[1,1,1]
    @test_throws ErrorException A[-1]
    
    B = ShapeOnlyArray{Int8,1}((3,))
    @test size(B) == (3,)
    @test eltype(B) == Int8
    for i in 0:4
        @test_throws ErrorException B[i]
    end
    
    C = ShapeOnlyArray{Bool,2}((0x2,UInt16(4095)))
    @test size(C) === (2,4095)
    @test eltype(C) == Bool
end

```

Do some unit tests on ShapeOnlyArray?

test/runtests.jl
```diff
@@ -256,6 +256,17 @@ end
...
 
+@testset "zcreate does not allocate dense storage" begin
+    mktempdir() do dir
+        # Tens of TB if zcreate were materializing dummy storage.
+        r = @timed zcreate(UInt8, 10674, 10653, 9327; path = joinpath(dir, "big.zarr"))
+        @test r.bytes < 1e9
```
mkitti, line 283: This should be much less than 1e9?

test/runtests.jl
mkitti: This will address the coverage issue.

src/ZArray.jl
```diff
@@ -339,6 +339,13 @@ function zcreate(::Type{T}, dims::Integer...;
   zcreate(T, store, dims...; zarr_format, dimension_separator, kwargs...)
 end
 
+struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
+    sz::Dims{N}
+end
```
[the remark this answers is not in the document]
mkitti: This leads to a stackoverflow error.

mkitti, line 344:
```suggestion
struct ShapeOnlyArray{T,N} <: AbstractArray{T,N}
    sz::Dims{N}
    ShapeOnlyArray{T,N}(dims::Dims{N}) where {T,N} = new{T,N}(dims)
end
```

This will help with Copilot's suggestion and avoid the StackOverlfow.

bjarthur:
"Starting April 20, 2026, new sign-ups for Copilot Pro, Copilot Pro+, and student plans are temporarily paused."

src/ZArray.jl
bjarthur: had to remove this line to get `ShapeOnlyArray{Bool,2}((0x2,UInt16(4095)))` to work

test/runtests.jl
bjarthur: by 4x yes, but why push the test close to where it might fail?

```
julia> r.bytes
242486544
```

src/ZArray.jl
bjarthur: covered now

test/runtests.jl
mkitti: Where are all the allocations going?

src/ZArray.jl
mkitti: My line was only needed if we also added Copilot's suggestion. I forgot that constructors also do some argument conversion.

mkitti (approved):
I'm still concerned about the number of allocations occurring here, but we will explore later.

bjarthur:
`julia --track-allocation` doesn't return much in Zarr.jl/src, but it seems to be in compilation somewhere as running `zcreate` a second time results in much less allocation:

```
julia> using Zarr

julia> dir = mktempdir()
"/Users/arthurb/.tmp/jl_VYeH3E"

julia> r = @timed zcreate(UInt8, 10674, 10653, 9327; path = joinpath(dir, "big3.zarr"))
(value = ZArray{UInt8} of size 10674 x 10653 x 9327, time = 0.641878209, bytes = 242482944, gctime = 0.021386041, gcstats = Base.GC_Diff(242482944, 4611, 0, 4700977, 19, 3253, 21386041, 3, 0), lock_conflicts = 0, compile_time = 0.640550542, recompile_time = 0.0)

julia> r = @timed zcreate(UInt8, 10674, 10653, 9327; path = joinpath(dir, "big4.zarr"))
(value = ZArray{UInt8} of size 10674 x 10653 x 9327, time = 0.001134458, bytes = 16384, gctime = 0.0, gcstats = Base.GC_Diff(16384, 0, 0, 233, 0, 0, 0, 0, 0), lock_conflicts = 0, compile_time = 0.0, recompile_time = 0.0)
```

mkitti:
Sounds like we should work on precompilation.

Issue #267, opened 2026-05-05
getattrs for version 3


lazarusA:
current function goes looking for `.zattrs`, it shouldn't. It needs to look for `.json`.

mentioned in #266

Issue #266, opened 2026-04-28
Support dimension_separator


orenbenkiki:
Using "/" would allow using M directories w/ N files per directory instead of M*N files in a single directory, which matters for large matrices. This is allowed by the Zarr spec. Arguably "/" would have been a better default than ".", but at least allow overriding it.

lazarusA:
Hi,

could you share more details of what is not working? We have tests for that case (I think): https://github.com/JuliaIO/Zarr.jl/blob/56eda22ff244d097545016eb68d1b9c19b141dd1/test/v3_julia.jl#L27

so, I'm not sure I follow fully.

orenbenkiki:
My bad. I was using 0.9.6 - this is fixed in 0.10. That said... what I really want to use is the shard file (and not one file per chunk on disk). Which means using V3.  It reorganized the pipeline (array→array → array→bytes → bytes→bytes codec slots to replace v2's filters  → compressor), and Zarr.jl 0.10's v3 codec registry is sparse. Three concrete gaps that bite me:

  1. No vlen-utf8 v3 codec. v2 had VLenUTF8Filter auto-applied for String arrays. v3's BytesCodec is the default array→bytes codec and refuses String input. → I generated a local VLenUTF8V3Codec (see below).
  2. No standalone shuffle codec. v2 has ShuffleFilter (used by :gzip_shuffle). v3 has no shuffle codec, only the shuffle setting bundled inside BloscV3Codec. → :gzip_shuffle errors on my v3 backend.
  3. No standalone bitshuffle codec. Same story — bitshuffle is only available inside blosc. → :zstd_bitshuffle errors (and was already rejected in v2 too, just for a different reason).

I assume this will all get fixed/implemented eventually... thanks for doing this project!

```julia
# ==============================================================================
# BEGIN: temporary `vlen-utf8` v3 codec.
#
# Zarr.jl 0.10's V3 codec registry has no `vlen-utf8` entry, and its v3 metadata
# constructor silently drops the `filters` kwarg, so writing a Zarr v3 array of
# `String` fails (the default `BytesCodec` cannot reinterpret `String`). The
# codec below mirrors the wire format of Zarr.jl's v2 `VLenUTF8Filter` (numcodecs
# `vlen-utf8`) so v3 stores written here round-trip through the same standard
# encoding. The matching `string_zcreate` helper bypasses Zarr.jl's high-level
# `zcreate` to construct `MetadataV3` with this codec in the pipeline.
#
[...]

PR #270, opened 2026-05-08
getattrs and consolidate metadata v3

### LINKED ISSUE #268: Support Python-compatible v3 metadata
Until ZEP-309 is finalized, be compatible with Python for V3 metadata (split out of #266; relevant for #257). Provide similar API to V2 (rebuilding consolidated metadata etc.).

lazarusA:
closes #268 #267 

following the suggestions from https://github.com/JuliaIO/Zarr.jl/issues/266#issuecomment-4356178032

### DIFF
--- a/src/Storage/consolidated.jl
+++ b/src/Storage/consolidated.jl
@@ -9,12 +9,22 @@ struct ConsolidatedStore{P} <: AbstractStore
   cons::Dict{String,Any}
 end
 function ConsolidatedStore(s::AbstractStore, p)
+  # v2-style: separate .zmetadata file
   d = s[p, ".zmetadata"]
-  if d === nothing
-    throw(ArgumentError("Could not find consolidated metadata for store $s"))
+  if d !== nothing
+      meta = JSON.parse(String(copy(d)); dicttype = Dict{String,Any})
+      return ConsolidatedStore(s, p, meta["metadata"])
   end
-  meta = JSON.parse(String(copy(d)); dicttype = Dict{String,Any})
-  ConsolidatedStore(s, p, meta["metadata"])
+  # v3-style: consolidated_metadata embedded inside root zarr.json
+  z = s[p, "zarr.json"]
+  if z !== nothing
+      root = JSON.parse(String(copy(z)); dicttype = Dict{String,Any})
+      cm = get(root, "consolidated_metadata", nothing)
+      if cm !== nothing
+          return ConsolidatedStore(s, p, cm["metadata"])
+      end
+  end
+  throw(ArgumentError("Could not find consolidated metadata for store $s"))
 end
 
 function Base.show(io::IO,d::ConsolidatedStore)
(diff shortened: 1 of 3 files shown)


src/Storage/consolidated.jl
```diff
@@ -38,8 +48,20 @@ function getattrs(::ZarrFormat{2}, d::ConsolidatedStore, p)
...
+  node_meta = get(d.cons, json_key, nothing)
+  if node_meta === nothing
+      return Dict{String,Any}()
+  end
+  # v3 spec: user attributes are stored under the "attributes" subkey of zarr.json
+  if haskey(node_meta, "attributes")
+      return node_meta["attributes"]
+  end
+  # fallback: return the full node metadata if no "attributes" key is found, we never write this. We only read it.
+  # is this correct?
```
mkitti, line 61: My perspective here as a late mover on this is that the top-level consolidated metadata key has no basis under the Zarr v3 specification.

The one extension mechanism that we do have is Zarr conventions.

https://github.com/zarr-conventions

I think we should create and use a Zarr convention for consolidated metadata under attributes.

Effectively though, this would create two different consolidated metadata stores.

A third route is suggested by the structure here. Perhaps consolidated metadata is actually a storage transformer. I think this is actually complementary to the Zarr convention.

For now, I think we actually have two stores:
1. CompatibleConsolidatedMetadata - implements the top-level key
2. ConventionalConsolidatedMetadata - implements the attributes subkey, zarr convention declaration, and Storage Transformer extension.

We should be able to read and write both.

mkitti (commented):
I think we should pursue dual tracks here. One that can read and write the de facto consolidated metadata as a top-level key and another that pursues a zarr convention and extension approach.

src/Storage/consolidated.jl
lazarusA: but, writing is not a fluke? like, it should not be encouraged to allow the top-level. Hence, only reading will do?
Also, you are saying that we should considering incorporating 2 new structs for splitting these scenarios?

mkitti: Eventually, we should only write under attributes. However, no one else can read that at the moment. We need consensus on the convention.

lazarusA: I see. You have a better mental model of what is going on, probably it would be better if you could finish this PR? if time is on you side.

mkitti: I will focus on getting this merged for now.

mkitti:
So for now, I think we should take out the consolidated metadata under attributes thing. I will deal with that in a separate pull request.

This pull request should just deal with consolidated metadata as a top-level key. @lazarusA , could you finish this under those parameters?

lazarusA:
like in here #277  ? I will close this one. I opened https://github.com/JuliaIO/Zarr.jl/issues/279 to keep track of the remaining issues.

lazarusA:
doing just the `consolidated metadata as a top-level key` in #287. Decided to do it there with a different approach.

PR #280, opened 2026-05-20
Performance improvements (v2 only for now)

asinghvi17:
This PR works through @glwagner's list of V2 performance issues and lands the biggest wins as separate commits. V3 is out of scope here (covered separately).

## Commits

| | sha | summary |
|---|---|---|
| 1 | `59644e4` | A1 — bulk-copy `zcompress!` fallback |
| 2 | `50fbc16` | B1 — bulk-copy `NoCompressor` `zuncompress!` |
| 3 | `c046338` | A2 — `getchunkarray_undef` skips dead zero-fill |
| 4 | `ebc3ea1` | A3-write — `writeblock!` single-chunk fast path |
| 5 | `d846986` | A3-read — `readblock!` single-chunk fast path |
| 6 | `3839b95` | factor `singlechunk_fastpath` guard helper |

### A1 — `zcompress!` fallback (`59644e4`)

Replace `empty!` + `append!` with `resize!` + `copyto!` in the generic `zcompress!` fallback in `Compressors.jl`. For `NoCompressor`, `zcompress` returns a lazy `reinterpret(UInt8, data)` view that the old path walked element-by-element through `_growend!` / `push!` — about 67% of CPU on uncompressed V2 write profiles. The new path is a single SIMD/memcpy bulk copy. Real compressors (Blosc/Zstd/Zlib) already return a `Vector{UInt8}` and `copyto!` handles them identically.

### B1 — `NoCompressor` `zuncompress!` (`50fbc16`)

Mirror image for the decode side. The generic fallback ends up at `copyto!(::Array{T}, ::ReinterpretArray)`, which walks element-by-element at ~17 GB/s on Apple silicon vs. ~85 GB/s for `unsafe_copyto!` on the same memory. Add a `::NoCompressor`-dispatched `zuncompress!` method alongside the existing per-compressor versions (Zstd / Zlib / Blosc already had their own). Guards on `data::Array{T}` and `isbitstype(T)` keep the fast path off `SenMissArray`, `MaxLengthString`, ragged `Vector{T}` eltypes, and non-contiguous inputs.

### A2 — `getchunkarray_undef` (`c046338`)

Both `readblock!` and `writeblock!` allocated their chunk-shaped scratch buffer via `getchunkarray(z) = fill(_zero(eltype(z)), z.metadata.chunks)`, then immediately overwrote it (decode writes every element on reads; `resetbuffer!` re-initialises on writes). Add `getchunkarray_undef(z::ZArray{T})` returning `Array{T}(undef, chunks)`, used by `readblock!` always and by `writeblock!` when `fill_value !== nothing`. Falls back to the zero-fill when `Missing <: T` (the `SenMissArray` codec path needs the isbits inner buffer; Blosc rejects `Union{Missing,T}` directly). Port of the V2-side hunks from #272 verbatim, including the `Missing <: T` carve-out that PR added in `75bea38` to fix a Blosc rejection regression.

### A3-write — `writeblock!` fast path (`ebc3ea1`)

PProf of A1+B1+A2 showed 93% of write CPU pinned in `__psynch_cvwait` — the readtask/writetask 0-buffered channel ferry sync wait. For the dominant single-chunk full-overwrite case (`z[:,:,:,t] = buf`), add a fast path that bypasses the channels entirely: encode `ain` straight through `compress_raw` and call `store_writechunk` / `store_deletechunk` synchronously. Fill-value elision is preserved. Anything else falls through to the channel-based path. Skips `Missing <: T` so the `SenMissArray` codec contract stays intact.

### A3-read — `readblock!` fast path (`d846986`)

Same idea on the read side. Profile showed 80% read CPU in `__psynch_cvwait` plus 13% GC pressure from the per-call chunk-shaped scratch + final copy. The fast path reads the compressed bytes synchronously via `store_readchunk` and decodes directly into `aout`, eliminating both the channel ferry and the scratch buffer.

### Cleanup — `singlechunk_fastpath` helper (`3839b95`)

Both fast paths share the same eligibility check. Factor it into `singlechunk_fastpath(arr, z, blockr) -> Union{CartesianIndex, Nothing}` so each caller's fast-path body is just the I/O. Also drops the redundant `indranges` coverage check — when `size(arr) == chunks` and `length(blockr) == 1`, the slice must align with that chunk's range (otherwise it would straddle into a neighbouring chunk and bump `length(blockr)`).

## Combined effect — master vs HEAD

Apple M2 Ultra, Julia 1.12.6, V2 + NoCompressor + `chunks=(Nx,Ny,Nz,1)`, 1 warm-up + 7 timed repeats per size, matched A/B (`git checkout 03270b2 -- src/` then back) on the same wall-clock so cache state and power profile are identical:

### Writes

| size | master (03270b2) | this PR | speed-up |
|---|---:|---:|---:|
| 128×128×16×50 (50 MiB)   |  710 MB/s | 2172 MB/s | **3.06×** |
| 256×256×32×50 (400 MiB)  |  740 MB/s | 2313 MB/s | **3.12×** |
| 512×512×32×50 (1.56 GiB) |  757 MB/s | 2184 MB/s | **2.88×** |

### Reads

| size | master (03270b2) | this PR | speed-up |
|---|---:|---:|---:|
| 128×128×16×50 (50 MiB)   | 3261 MB/s | 7061 MB/s | **2.16×** |
| 256×256×32×50 (400 MiB)  | 3008 MB/s | 6746 MB/s | **2.24×** |
| 512×512×32×50 (1.56 GiB) | 3304 MB/s | 8010 MB/s | **2.42×** |

### vs Zarrs.jl

For reference, Zarrs.jl (the Rust-backed implementation) on the same workload:

| size | this PR (W) | Zarrs (W) | this PR (R) | Zarrs (R) |
|---|---:|---:|---:|---:|
| 50 MiB  | 2172 | 1109 | 7061 | 5435 |
| 400 MiB | 2313 | 1889 | 6746 | 5053 |
| 1.5 GiB | 2184 | 1991 | 8010 | 3697 |

Zarr.jl is now **faster than Zarrs.jl** on uncompressed V2 across all three sizes, on both writes (1.10–1.96×) and reads (1.30–2.17×).

## Tests

2499/2499 pass after each commit. Paths exercised:

- `SenMissArray` / `Missing <: T` — `Fillvalue as missing` and the `getindex/setindex` amiss tests
- `MaxLengthString large-chunk read path`
- Ragged arrays (`Vector{Float64}` eltype)
- Python round-trip via PythonCall
- V3 fill-value elision (`v3_codecs.jl:336`) — V3 path is unchanged

## Not in this PR

- **V3 path.** `BytesCodec` encode/decode bulk-copy and the exact-full-chunk overwrite path are still open; covered by upstream #272 (V3 portion) and a separate write-up.
- **A5** — thread a scratch `Vector{UInt8}` through `pipeline_encode` (main benefit is for compressed codecs, not NoCompressor).
- **A6** — `DirectoryStore` `mkdir` cache (bigger win for many-small-chunks workloads).
- **Dropping `all(isequal(fill_value), data)`.** Confirmed no measurable win on dense bench data (`all` short-circuits at element 1 in nanoseconds) and has real semantic cost: it implements Zarr's fill-value chunk elision, which would change the on-disk shape for sparse arrays and re-writes-with-fill, plus break 3 tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- V2 writes: add `writeblock!` fast path for single-chunk full-overwrites that bypasses the readtask/writetask channels and the chunk-shaped scratch buffer, encoding straight from the input array into the store [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
 - V2 NoCompressor writes: replace `append!` over the reinterpret view with bulk `resize!` + `copyto!` in the generic `zcompress!` fallback [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
 - V2 NoCompressor reads: add bulk-copy `zuncompress!` method dispatched on `::NoCompressor` to bypass `copyto!(::Array, ::ReinterpretArray)`'s element-by-element walk [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
 - V2 read+write chunk allocation: add `getchunkarray_undef` and skip the dead zero-fill of the chunk-shaped scratch buffer on full-overwrite paths [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- V2 reads: add `readblock!` fast path for single-chunk full-reads that bypasses the readtask channel and the chunk-shaped scratch buffer, decoding straight into the output array [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
 - V2 writes: add `writeblock!` fast path for single-chunk full-overwrites that bypasses the readtask/writetask channels and the chunk-shaped scratch buffer, encoding straight from the input array into the store [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
 - V2 NoCompressor writes: replace `append!` over the reinterpret view with bulk `resize!` + `copyto!` in the generic `zcompress!` fallback [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
 - V2 NoCompressor reads: add bulk-copy `zuncompress!` method dispatched on `::NoCompressor` to bypass `copyto!(::Array, ::ReinterpretArray)`'s element-by-element walk [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 ## Unreleased
 
-- V2 reads: add `readblock!` fast path for single-chunk full-reads that bypasses the readtask channel and the chunk-shaped scratch buffer, decoding straight into the output array [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
-- V2 writes: add `writeblock!` fast path for single-chunk full-overwrites that bypasses the readtask/writetask channels and the chunk-shaped scratch buffer, encoding straight from the input array into the store [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
-- V2 NoCompressor writes: replace `append!` over the reinterpret view with bulk `resize!` + `copyto!` in the generic `zcompress!` fallback [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
-- V2 NoCompressor reads: add bulk-copy `zuncompress!` method dispatched on `::NoCompressor` to bypass `copyto!(::Array, ::ReinterpretArray)`'s element-by-element walk [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- V2 NoCompressor writes: replace `append!` over the reinterpret view with bulk `resize!` + `copyto!` in the generic `zcompress!` fallback [#280](https://github.com/JuliaIO/Zarr.jl/pull/280)
```
lazarusA: maybe clean up also this one. Only one PR number and link is enough. Additional added features can go afterwards as a list.

asinghvi17: Done in 031bf40 — collapsed into one [#280] entry with sub-bullets matching the v0.10.0 #241 style.

lazarusA:
There is still a lot of AI slop in between functions, maybe another pass to cleanup ? or change them to docstrings (making sure the statements are accurate) if you want to keep them. Other than that, it does the job!

lazarusA (approved):

lazarusA:
looking forward to the v3 improvements 😄 . I will merge now, and if there are issues we fix them later in smaller PRs.

PR #293, opened 2026-06-05
Add another caching approach

meggart:
Following the discussion around #231 I added a Zarrcache function which implements another array-based permanent cache that makes use of the DiskArrays CachedDiskArray type. In order to clarify potential use cases, I have added a section to the docs User Guide to descrube when to use which cache.


Project.toml
```diff
@@ -36,8 +36,8 @@ ChunkCodecLibZstd = "1"
 CRC32c = "1.10, 1.11"
 DataStructures = "0.17, 0.18, 0.19"
 DateTimes64 = "1"
-DiskArrays = "0.4.2"
-HTTP = "^1.3, 2"
+DiskArrays = "0.4.21"
+HTTP = "^1.3"
```
lazarusA, line 40: could we keep also  `HTTP 2`?

meggart: No, because tests did not pass locally for me with HTTP 2

meggart: Something is really strange. When I run the Zarr tests with Pkg.test() they pas for me locally but running these lines manually, which are part of the test throws an error:

````julia
  using Zarr.HTTP, Sockets
  server = Sockets.listen(0)
  ip, port = getsockname(server)
  @async HTTP.serve(g, ip, port, server=server)
````

Do you have an idea what is going on?

Comment (lazarusA) [reply Project.toml:40]: for HTTP 2, `server=server` is not longer allowed. And using `HTTP.serve!(handler, host, port; kwargs...)` is the current recommendation, it takes care of all ports life-cycle.

Comment (lazarusA) [reply Project.toml:40]: I did this in Browzarr.jl

https://github.com/EarthyScience/Browzarr.jl/blob/e5acc145d6484ca74d28c297bf02895c83fb5361/src/Browzarr.jl#L7

```Julia
using Zarr, Sockets
using Zarr: DirectoryStore

# compat abstractions between HTTP 1.x and 2.x
const HTTP_V2 = pkgversion(HTTP) >= v"2.0"
const server_ = HTTP_V2 ? HTTP.Server : Sockets.TCPServer

function _serve!(handler, host, port; kwargs...)
    if HTTP_V2
        return HTTP.serve!(handler, host, port; kwargs...)
    else
        server = Sockets.listen(Sockets.getaddrinfo(host), port)
        return server, errormonitor(@async HTTP.serve(handler, host, port, server = server))
    end
end

_get_server(s) = HTTP_V2 ? s : first(s)

function _port(s)
    if HTTP_V2
        return HTTP.port(s)
    else
        _, port = getsockname(first(s))
        return port
    end
end

```

meggart: Can we fix the `server=server` deprecation in a separate PR? It is not really related to my changes.

meggart: It is just obvious that we currently do not support HTTP , so we should not claim we do.

Issue #275, opened 2026-05-19
`writeblock!` returns before its async chunk writetask completes → caller can observe truncated store


tomerarnon:
`src/ZArray.jl:187-244`: `writeblock!` spawns a fire-and-forget chunk-write task and returns when the channel-feed `for` loop completes:

```julia
writetask = @async write_items!(z.storage, writechannel, z.path, blockr)
bind(writechannel, writetask)
try
    for i in 1:length(blockr)
        # ... compress + put! onto writechannel
    end
finally
    close(readchannel)
    close(writechannel)   # ← function returns here; writetask may still be running
end
```

The caller sees `arr .= data` return successfully, but on stores where `setindex!` is genuinely async (notably `S3Store`), the underlying HTTPS PUT may not have landed yet. Downstream observers — `subkeys`, `isinitialized`, or a separate process reading the store — see truncated state for an unbounded window after the write call returned.

### Observed impact

In a workload writing many arrays in close succession on `S3Store` and verifying chunk presence immediately afterward, we saw 17–49 "missing" arrays per export pass. 97% resolved within ~100 ms; the rest within ~300 ms. A 2 s drain before verifying eliminated the issue entirely — confirming the writes do land, just not by the time `writeblock!` returns. We worked around it downstream with a drain-then-verify sweep, but the lack of synchronization is the upstream defect.

### Secondary defect

If the writetask ever throws (S3 5xx, network blip post-handshake, OOM-killed task), the exception is swallowed by GC of an un-`wait`ed task. We have no direct evidence of this firing in production, but the code path is open.

### Suggested fix

```julia
finally
    close(readchannel)
    close(writechannel)
    wait(writetask)          # synchronize; also propagates any exception
end
```

Same change applies symmetrically to `readtask` if it can throw post-handshake.

### Test

A `TestStore` whose `setindex!` calls `sleep(0.1)` before assigning the buffer demonstrates that today's `writeblock!` returns before the chunk is observable via `subkeys`; the patch makes it observable.

### Related

#271 (`subkeys(::S3Store)` silent truncation above 1000 keys). Both are symptoms of fire-and-forget tasks in the storage layer.

PR #297, opened 2026-06-17
fix: write chunks in the ConcurrentRead write_items! path

tomerarnon:
## Problem

`write_items!` for the `ConcurrentRead` strategy silently drops every chunk write. Stores that advertise `ConcurrentRead` — notably `S3Store` — end up with array metadata (`.zarray`/`.zattrs`) written but **no data chunks**, and no error is raised.

The cause is a one-character typo in `src/Storage/Storage.jl`. The `ConcurrentRead` method writes a chunk with:

```julia
store_writechunk(s, data, p, ii, e) = data
```

The `= data` makes this a **no-op local function definition** (short-form `f(args) = body`) instead of a call, so `store_writechunk` is never invoked and the chunk is never written. The `SequentialRead` method just above it calls it correctly:

```julia
store_writechunk(s, data, p, ii, e)
```

This is why `DirectoryStore`/`DictStore` (default `SequentialRead`) write correctly while `S3Store` (`ConcurrentRead`) produces metadata-only stores.

## Fix

Drop the `= data` so the chunk is actually written.

## Reproduction (no S3 needed)

A `DictStore` wrapper whose only difference is advertising `ConcurrentRead`:

```julia
using Zarr
import Zarr: AbstractStore, DictStore, ConcurrentRead, store_read_strategy, subdirs, subkeys

struct ConcurrentDictStore <: AbstractStore
    inner::DictStore
end
ConcurrentDictStore() = ConcurrentDictStore(DictStore())
Base.getindex(s::ConcurrentDictStore, i::AbstractString) = s.inner[i]
Base.setindex!(s::ConcurrentDictStore, v, i::AbstractString) = (s.inner[i] = v)
Base.delete!(s::ConcurrentDictStore, i::AbstractString) = delete!(s.inner, i)
subdirs(s::ConcurrentDictStore, p) = subdirs(s.inner, p)
subkeys(s::ConcurrentDictStore, p) = subkeys(s.inner, p)
Zarr.storagesize(s::ConcurrentDictStore, p) = Zarr.storagesize(s.inner, p)
store_read_strategy(::ConcurrentDictStore) = ConcurrentRead(4)

data = reshape(collect(Float32, 1:12), 3, 4)
store = ConcurrentDictStore()
a = zcreate(Float32, zgroup(store), "x", 3, 4; chunks=(3, 4))
a[:, :] = data

@assert haskey(store.inner.a, "x/0.0")  # fails on current main — chunk never written
```

On `main` the chunk key is absent; with this fix it is present and round-trips.

## Test

Adds a regression test in `test/storage.jl` using the wrapper above, asserting the chunk is written and round-trips.

## Note

This looks related to #275 / #281 (the `wait(writetask)` fix): that addressed write-task *timing*, but the chunk write in the `ConcurrentRead` path is a no-op, so waiting for it still writes nothing. The no-op is still present on `master`.

Regressed in #226 (the v3 rewrite, shipped in 0.10.0): the direct `s[p,ii] = data` write was refactored into a `store_writechunk` helper, and the `ConcurrentRead` branch's call was written as an assignment. `v0.9.6` wrote chunks correctly.

### DIFF
--- a/src/Storage/Storage.jl
+++ b/src/Storage/Storage.jl
@@ -278,7 +278,7 @@ function write_items!(s::AbstractStore, c::AbstractChannel, r::ConcurrentRead, e
         store_deletechunk(s, p, ii, e)
         end
       else
-      store_writechunk(s, data, p, ii, e) = data
+      store_writechunk(s, data, p, ii, e)
       end
       nothing
   end
--- a/test/storage.jl
+++ b/test/storage.jl
@@ -402,3 +402,27 @@ end
   # Cleanup
   rm(cache_dir, recursive=true)
 end
+
+# A store identical to DictStore but advertising the ConcurrentRead strategy
+# (as S3Store does), to exercise the concurrent write path.
+struct ConcurrentDictStore <: Zarr.AbstractStore
+  inner::Zarr.DictStore
+end
+ConcurrentDictStore() = ConcurrentDictStore(Zarr.DictStore())
+Base.getindex(s::ConcurrentDictStore, i::AbstractString) = s.inner[i]
+Base.setindex!(s::ConcurrentDictStore, v, i::AbstractString) = (s.inner[i] = v)
+Base.delete!(s::ConcurrentDictStore, i::AbstractString) = delete!(s.inner, i)
+Zarr.subdirs(s::ConcurrentDictStore, p) = Zarr.subdirs(s.inner, p)
+Zarr.subkeys(s::ConcurrentDictStore, p) = Zarr.subkeys(s.inner, p)
+Zarr.storagesize(s::ConcurrentDictStore, p) = Zarr.storagesize(s.inner, p)
+Zarr.store_read_strategy(::ConcurrentDictStore) = Zarr.ConcurrentRead(4)
+
+@testset "ConcurrentRead stores persist chunk writes" begin
+  data = reshape(collect(Float32, 1:12), 3, 4)
+  store = ConcurrentDictStore()
+  g = zgroup(store)
+  a = zcreate(Float32, g, "x", 3, 4; chunks=(3, 4))
+  a[:, :] = data
+  @test haskey(store.inner.a, "x/0.0")   # chunk data actually written, not just metadata
+  @test a[:, :] == data                  # round-trips through the store
+end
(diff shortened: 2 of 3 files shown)


src/Storage/Storage.jl
```diff
@@ -278,7 +278,7 @@ function write_items!(s::AbstractStore, c::AbstractChannel, r::ConcurrentRead, e
         store_deletechunk(s, p, ii, e)
         end
       else
-      store_writechunk(s, data, p, ii, e) = data
```
lazarusA, line 281: oh, yes, this looks like a bug. @meggart ?

test/storage.jl
```diff
@@ -402,3 +402,27 @@ end
...
+  inner::Zarr.DictStore
+end
+ConcurrentDictStore() = ConcurrentDictStore(Zarr.DictStore())
+Base.getindex(s::ConcurrentDictStore, i::AbstractString) = s.inner[i]
+Base.setindex!(s::ConcurrentDictStore, v, i::AbstractString) = (s.inner[i] = v)
+Base.delete!(s::ConcurrentDictStore, i::AbstractString) = delete!(s.inner, i)
+Zarr.subdirs(s::ConcurrentDictStore, p) = Zarr.subdirs(s.inner, p)
+Zarr.subkeys(s::ConcurrentDictStore, p) = Zarr.subkeys(s.inner, p)
+Zarr.storagesize(s::ConcurrentDictStore, p) = Zarr.storagesize(s.inner, p)
+Zarr.store_read_strategy(::ConcurrentDictStore) = Zarr.ConcurrentRead(4)
```
lazarusA, line 418: should we have this also in `Storage.jl`, or? not sure.

lazarusA:
please add an entry to the changelog.

test/storage.jl
tomerarnon: I put this together just for the sake of having a MWE that wasn't tied to any of my real usage code, but I'm happy to move this if you'd like (or track in a follow on issue)

lazarusA: is also in the PR description for future reference. I think we are good for now. However, I would like to have this bit in a more permanent place, somehow, maybe the docs. But, we do that later.

lazarusA (approved):

Issue #310, opened 2026-07-27
zopen(url; consolidated=true) throws MethodError for cloud stores: ConsolidatedStore path field rejects the SubString returned by storefromstring


haavardhvarnes:
### Bug

`zopen(url; consolidated=true)` throws a `MethodError` for cloud stores (GCStore, HTTPStore) in Zarr.jl v0.10.1:

```julia
julia> using Zarr
julia> zopen("gs://gcp-public-data-arco-era5/ar/full_37-1h-0p25deg-chunk-1.zarr-v3"; consolidated=true)
ERROR: MethodError: no method matching Zarr.ConsolidatedStore(::Zarr.GCStore, ::SubString{String}, ::Dict{String, Any})
```

(The store is public/anonymous, so this reproduces as-is; opening the same group non-consolidated, or any local `DirectoryStore` path with `consolidated=true`, works fine.)

### Root cause

- `storefromstring(::Type{<:GCStore}, url, _)` (`src/Storage/gcstore.jl:137`) produces the path via `lstrip(uri.path, '/')` (or `split(...)[2]`), i.e. a `SubString{String}`. The HTTPStore method has the same property.
- `ConsolidatedStore` is a parametric struct with field `path::String` (`src/Storage/consolidated.jl:6-10`). Its implicit outer constructor therefore only accepts `::String` exactly — no convert — so the 3-arg call `ConsolidatedStore(s, p::SubString, metadata::Dict)` reached from `ConsolidatedStore(s, p, ::ZarrFormat{2})` (`consolidated.jl:12-18`) has no matching method.
- `DirectoryStore` paths come back as `String`, which is why the bug only bites remote stores.

### Suggested fix

Either convert at the format-dispatch constructors (`return ConsolidatedStore(s, String(p), metadata)` in the `ZarrFormat{2}`/`ZarrFormat{3}` methods), or add

```julia
ConsolidatedStore(p::AbstractStore, path::AbstractString, cons::AbstractDict{String,Any}) =
    ConsolidatedStore(p, String(path), cons)
```

### Workaround

```julia
s, p = Zarr.storefromstring(url)
g = zopen(Zarr.ConsolidatedStore(s, String(p)), "r"; path = String(p))
```

(Note the `path =` kwarg is also required — constructing the `ConsolidatedStore` alone is not enough.)

[...]

Issue #313, opened 2026-08-05
HTTPStore fallback crashes with MethodError: no method matching subdirs(::HTTPStore, ::String) under HTTP.jl 2.x


dr-ko:
`HTTPStore` fallback path crashes with `MethodError: no method matching subdirs(::Zarr.HTTPStore, ::String)` under HTTP.jl 2.x

## Summary

Opening a remote Zarr store over plain HTTP(S) works with `HTTP.jl` 1.x but throws `MethodError: no method matching subdirs(::Zarr.HTTPStore, ::String)` with `HTTP.jl` 2.x. Downgrading to `HTTP@1` fixes it with no other code changes.

## Root cause

`storefromstring` for `HTTPStore` (`src/Storage/http.jl`) tries to wrap the store in a `ConsolidatedStore`, and on any failure falls back to returning the *raw* `HTTPStore`:

```julia
function storefromstring(::Type{<:HTTPStore}, s,_)
    http_store = HTTPStore(s)
    try
        cs = ConsolidatedStore(http_store, "")
        return cs, ""
    catch err
        @warn exception=err "Additional metadata was not available for HTTPStore."
    end
    return http_store,""
end
```

When that raw `HTTPStore` is later passed into `ZGroup` (`src/ZGroup.jl`), it unconditionally calls `subdirs(s, path)`:

```julia
function ZGroup(s::T, mode="r", path="", zarr_format=:auto; fill_as_missing=false) where T<:AbstractStore
  ...
  for d in subdirs(s,path)
```

`subdirs` is implemented for `DirectoryStore`, `DictStore`, and `GCStore`, but there is no method for plain `HTTPStore` — so any code path that reaches this fallback (i.e. whenever `ConsolidatedStore(http_store, "")` throws) hits a hard `MethodError` instead of a clean, catchable error.

## Why this only shows up with HTTP.jl 2.x

`HTTPStore.getindex` explicitly requests the OpenSSL TLS backend:

```julia
r = HTTP.request("GET", string(s.url,"/",k), status_exception = false, socket_type_tls=OpenSSL.SSLStream)
```

[...]

PR #314, opened 2026-08-08
Convert path to String in ConsolidatedStore constructors

rafaqz:
`ConsolidatedStore` needs `String` (e.g. not `SubString`) so we need to force the conversion.


lazarusA:
is this enough to close https://github.com/JuliaIO/Zarr.jl/issues/310 ? and partially #313 ....

rafaqz:
Should be, thats the error I was getting too

rafaqz:
Can we merge this? a tiny bugfix doesn't need a changelog

PR #317, opened 2026-08-17
ZarrCore.jl

asinghvi17:
- Move things around


felixcremer:
AbstractStore  public
ChunkKeyEncoding public
normalize_path private
VLenArrayFilter public
Every Codec, Filter and Compressor and Store is public
MaxLengthString private
ShapeOnlyArray private (should be removed)
isinitialized public (another API should be made in DiskArrays)
fill_value_decoding public
Metadata private
typestr public
ZarrFormat private
is_zgroup private
is_zarray private
zname public

felixcremer:
Why did the code coverage drop?

felixcremer:
Why does only ubuntu lts fail with the following error?

```julia

  Data type encoding: Error During Test at /home/runner/work/Zarr.jl/Zarr.jl/test/runtests.jl:269
    Test threw exception
    Expression: Zarr.typestr(Bool) === "|b1"
    UndefVarError: `typestr` not defined
    Stacktrace:
```

asinghvi17:
It's probably lts (julia 1.10), not ubuntu specifically.

src/Zarr.jl
```diff
@@ -1,33 +1,28 @@
...
+end
 
-export ZArray, ZGroup, zopen, zzeros, zcreate, storagesize, storageratio,
-  zinfo, DirectoryStore, S3Store, GCStore, zgroup
+@static if VERSION >= v"1.11"
+    include("public_names_zarr.jl")
+else
+    # For Julia 1.10, we have to parse the public names from the source file, since
+    # `public` is not a keyword and `names(ZarrCore)` only returns exported names.
+    let public_names = read(joinpath(@__DIR__, "..", "lib", "ZarrCore", "public_names_core.jl"), String)
```
felixcremer:
```suggestion
    let public_names = read(joinpath(@__DIR__, "..", "lib", "ZarrCore", "src", "public_names_core.jl"), String)
```

src/Zarr.jl
```diff
@@ -1,33 +1,28 @@
...
-  zinfo, DirectoryStore, S3Store, GCStore, zgroup
+@static if VERSION >= v"1.11"
+    include("public_names_zarr.jl")
+else
+    # For Julia 1.10, we have to parse the public names from the source file, since
+    # `public` is not a keyword and `names(ZarrCore)` only returns exported names.
+    let public_names = read(joinpath(@__DIR__, "..", "lib", "ZarrCore", "src", "public_names_core.jl"), String)
+        public_names = replace(public_names, "public" => "using ZarrCore: ")
+        eval(Meta.parseall(public_names))
+    end
```
mkitti, line 25: This currently imports all public symbols of ZarrCore.jl from ZarrCore.jl into Zarr.j for Julia 1.10. Originally, it imported all public symbols of Zarr.jl from ZarrCore.jl.

I believe we still want to do what I originally did. Hoervern this currently works since the public symbols of ZarrCore.jl is a strict superset of those of Zarr.jl.

mkitti:
My current inclination is to just merge this, and then fix further issues later.

src/Zarr.jl
felixcremer: I think, that this is a change only for 1.10 and hopefully won't matter anymore soon. 
1.11 is also importing all  the public symbols from ZarrCore into Zarr but they are protected by the machinery that builds on top of the public declaration.

lazarusA (approved):

PR #320, opened 2026-08-17
Improve show method for S3Store

felixcremer:
This shows the url and the bucket of an S3Store to the user.


ext/s3store.jl
```diff
@@ -41,6 +45,7 @@ function Zarr.storagesize(s::S3Store,p)
 end
 
 function Zarr.isinitialized(s::S3Store, i::String)
+  @show s, i
```
felixcremer:
```suggestion
```

lazarusA (approved):

PR #324, opened 2026-08-18
[WIP] Subpackage split

asinghvi17:
Splits `ZarrCore` into eight packages so it's cheap to depend on: **18 deps → 7** (four stdlib), **185 ms → 58 ms** load time. `ZarrCore` alone still reads and writes both v2 and v3.

`Zarr` hard-depends on all of them and re-exports everything, so `using Zarr` is behaviourally unchanged.

**New packages** (`lib/`): `ZarrBlosc`, `ZarrZlib`, `ZarrZstd` — each owns its v2 compressor *and* its matching v3 codec, since the v3 codecs delegate to the v2 implementation — plus `ZarrHTTP`, `ZarrGCS`, `ZarrZip`, and `ZarrS3` (owns `S3Store`; AWSS3 stays a weak dep, moved down from the umbrella).

**Main constraint:** a subpackage mutating another package's registry at top level has the mutation *silently discarded* at precompilation — no error, no warning. All registrations moved into `__init__` and were verified in fresh processes.

**Also fixed / cleaned up**
- `cloud_list_objects` and `missing_chunk_return_code!` were never declared in `ZarrCore` despite having methods on both sides of the new boundary
- `gcstore.jl` used `HTTP` without importing it, relying on `http.jl` leaking the binding
- dropped dead deps (`Pkg`, `DataStructures`, `OffsetArrays`) and dead public API (`BloscCodec`, `GzipCodec`)
- v2→v3 compressor mapping is now an extensible `v2_to_v3_codecs` generic instead of a hardcoded `if/elseif`
- store lookup is specificity-sorted, so it no longer depends on package load order
- CI `Pkg.develop`s each subpackage explicitly, since Julia 1.10 ignores `[sources]`/`[workspace]`

**Tests:** 2672 passed, 0 failed, 2 errored — identical to the pre-change baseline; both errors are pre-existing and environmental (Minio throttle, a Python `NoneType` in the consolidated-metadata roundtrip).

https://claude.ai/code/session_01SVtC9c84LqrDk57a6q1tSD


asinghvi17:
Closing in favor of #330

PR #330, opened 2026-08-21
Split Zarr backends into subpackages

asinghvi17:
## Changes

- Move HTTP, GCS, S3, and ZIP backends into `ZarrHTTP`, `ZarrGCS`, `ZarrS3`, and `ZarrZip`.
- Move Blosc, zlib, and Zstandard implementations into `ZarrBlosc`, `ZarrZlib`, and `ZarrZstd`; each package provides its v2 compressor and v3 codec.
- Keep the existing `Zarr` API by loading and re-exporting all subpackages.
- Make `ZarrCore.default_compressor()` return `NoCompressor()`; `ZarrBlosc` overrides it to return `BloscCompressor()`.
- Register compressors, codecs, and URL stores in `__init__`; match store URL patterns by specificity.
- Keep AWSS3 as a weak dependency of `ZarrS3`.
- Make `missing_chunk_return_code!` and `gcs_credentials` public; stop exposing `DateTime64`, `PermanentZarrCache`, `BloscCodec`, and `GzipCodec` through `Zarr`.

## Validation

- Full test suite: 2,799 passed.
- Bare-`ZarrCore` and `ZarrBlosc` default-compressor dispatch checks passed in fresh processes.
- Julia and Python v3 fixtures generated successfully.
- All workspace packages precompiled on Julia 1.12.6.
- Documentation build, tracked-file spelling checks, and `git diff --check` passed.


mkitti:
I believe the `__init__` functions are not very friendly to static compilation. We'll need to address that later, but for now let's put the contents of `__init__` into a separate function and then just call that function from `__init__`.

For static compilation, we might need to disable `__init__` and then call these registration functions explicitly somewhere outside of module load.

mkitti:
I am having mixed feelings about these `public_name_*.jl` files, perhaps they should just be `public_names.jl` and then we can just figure out what they apply to from their path. Let's leave as it in this pull request.

mkitti:
Perhaps we should change the authors line to the following:

```
authors = ["Fabian Gans <fgans@bgc-jena.mpg.de>", "Zarr.jl Contributors <zarr@julialang.org>"]
```

I can add the zarr@julialang.org as a group email address.

mkitti:
@nhz2 could you take a look at this way of breaking up Zarr.jl rather than using extensions?

asinghvi17:
Inits are generally fine so long as they are themselves type stable - which I haven't checked, to be fair.  These might not be though. I will have a look at that at some point in the next couple days to verify.

mkitti:
I loosely remember issues about `__init__` having unknown side effects, in this case, registration. For example, the order of plugin registration now depends on the order of module load. If there is an issue, then my design would be to create a Preferences.jl option to disable registration on `__init__` then allow people to call some function to manually register plugins.

asinghvi17:
The Preferences.jl thing sounds reasonable to have anyway.  Each module can provide a function `zarr_init!`, which can get called by the user and also invoked under a guard in the init function.

nhz2:
It is better to have all the packages at the same level like in https://github.com/MakieOrg/Makie.jl or https://github.com/JuliaIO/ChunkCodecs.jl
With the current layout, anyone who installs Zarr will get extra copies of all the packages in lib.

mkitti:
Hmm, that doesn't necessarily mean the packages need to be at the same level but in this case that perhaps Zarr.jl should be in a directory that is not the top directory level directory.

nhz2:
I feel like it is a bad idea to assume `__init__` will never be run concurrently, so the globals `codec_parsers` and `compressortypes` should be protected by locks.

It also doesn't make sense for this registration to go in `__init__`. The basic codecs most people use should have high quality implementations that are included and reliably just work for any installation. People using less common codecs probably don't need automatic registration on `using`.

mkitti:
To be clear, the current design is that Zarr.jl comes with all accessories included, meaning that all compressors and codecs are enabled and loaded.

For a more selective approach, a developer could instead depend on ZarrCore.jl and a select group of accessory packages.

nhz2:
It sounds like in that case Zarr should initialize these global tables and use them for high level convenience functions.
For example, the method `function zopen(s::String, mode="r"; kwargs...)` that makes use of the global `storageregexlist` could go in Zarr.jl while the lower level
`function zopen(s::AbstractStore, mode="r"; kwargs...)` could go in ZarrCore.

asinghvi17:
I don't think you can get around automatic registration if you want metadata parsing to work.   Someone who loads ZarrCore and ZarrBlosc should be able to load a local blosc-compressed store using plain zopen with no other issues, imo

lazarusA:
> It is better to have all the packages at the same level like in https://github.com/MakieOrg/Makie.jl or https://github.com/JuliaIO/ChunkCodecs.jl With the current layout, anyone who installs Zarr will get extra copies of all the packages in lib.

Indeed, and also probably Zarr should be in his own folder. Like the [SpeedyWeather](https://github.com/SpeedyWeather/SpeedyWeather.jl)  layout.

nhz2:
I think it is fine to have this kind of design, but some alternatives to consider are:
JLD2.jl: [type piracy on a Val{filter_id}](https://github.com/JuliaIO/JLD2.jl/blob/10252646a19d6668ca99762a1484c83d8da3ad54/filterpkgs/JLD2Lz4/src/JLD2Lz4.jl#L36)
Go zip package: [explicit register functions with common defaults preregistered](https://cs.opensource.google/go/go/+/refs/tags/go1.27.0:src/archive/zip/register.go;l=117-123) and [a per reader override](https://cs.opensource.google/go/go/+/refs/tags/go1.27.0:src/archive/zip/reader.go;l=179-195)

I think in either case the default compressors should be part of ZarrCore. The type piracy in the current design is pointing to that.
```
ZarrCore.default_compressor() = BloscCompressor()
```
And if we include blosc in core (which we should if it is actually the default) then we might as well have zlib and zstd too. Though I also think zstd or zstd with byte shuffle could be a better default because it avoid the current blosc thread safety and memory safety bugs.

mkitti:
> I think in either case the default compressors should be part of ZarrCore. The type piracy in the current design is pointing to that.\n\nZarrCore.default_compressor() = BloscCompressor()

My opinion here is that `ZarrCore` should not have a default compressor. Zarr.jl may want to implement a default conpressor based on the Zarr format.

zarr-python switched from blosc (v1) to zstd as the default compressor.

mkitti:
> JLD2.jl: type piracy on a Val{filter_id}

Where is the type piracy? JLD2 implements a method which it defines on `Val{filter_id}`. If it implemented a method for a function it does not own on a type it does not own, that would be type piracy.

nhz2:
The JLD2Lz4 package doesn't "own" `Val{32004}` (owned by Base) or `Filters.filtertype` (owned by JLD2), though I guess in the context of JLD2 filters you could say JLD2Lz4 owns `Val{32004}`.

kleinschmidt:
this is _awesome_ work y'all!!  is there a timeline for release?  I'm desperate to remove PythonCall and use the native julia stuff directly, and don't mind vendoring this stuff in the mean time if it's going to be a while...

asinghvi17:
I'm hoping in the next couple weeks

mkitti (approved):
Looks good.

Preferences.jl should be able to address any race condition registration issues should they arise.
We can address the default compressor later.

PR #333, opened 2026-08-26
Fix fill_as_missing for zarr v3 arrays

asinghvi17:
**Broken:** `fill_as_missing=true` on any zarr v3 array threw `ArgumentError: cannot reinterpret UInt8 as Union{Missing, Float64}` as soon as an initialized chunk was read. `lib/ZarrCore/src/pipeline.jl:45` (`pipeline_decode!(::V3Pipeline, ...)`) passed `eltype(output)` — the `Union{T,Missing}` of the `SenMissArray` — to `codec_decode`, which reinterprets raw bytes. The stored elements are plain `T`; the v2 path already strips `Missing` in `zuncompress`.

**Fix:** pass `Base.nonmissingtype(eltype(output))` instead (one line).

**Test:** `test/v3_codecs.jl`, testset `"fill_as_missing for v3 arrays"` — create a v3 array, write one chunk, reopen with `fill_as_missing=true`, check eltype, the written chunk, and that the untouched chunk reads as `missing`. Fails before, passes after.

Stacked on #330.

https://claude.ai/code/session_011Qme1Mr9UxwceM15pWBG2k

### DIFF
--- a/lib/ZarrCore/src/pipeline.jl
+++ b/lib/ZarrCore/src/pipeline.jl
@@ -42,7 +42,6 @@ function pipeline_decode!(p::V3Pipeline, output::AbstractArray, compressed::Vect
         (sz, codec) -> Codecs.V3Codecs.encoded_shape(codec, sz),
         p.array_array; init=size(output)
     )
-    # `fill_as_missing` output has eltype `Union{T,Missing}` but the stored elements are plain `T`
     arr = Codecs.V3Codecs.codec_decode(p.array_bytes, bytes, Base.nonmissingtype(eltype(output)), intermediate_shape; fill_value)
     # Phase 1 reverse: array->array codecs (reverse order)
     for codec in reverse(collect(p.array_array))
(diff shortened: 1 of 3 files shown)


lib/ZarrCore/src/pipeline.jl
```diff
@@ -42,7 +42,8 @@ function pipeline_decode!(p::V3Pipeline, output::AbstractArray, compressed::Vect
         (sz, codec) -> Codecs.V3Codecs.encoded_shape(codec, sz),
         p.array_array; init=size(output)
     )
-    arr = Codecs.V3Codecs.codec_decode(p.array_bytes, bytes, eltype(output), intermediate_shape; fill_value)
+    # `fill_as_missing` output has eltype `Union{T,Missing}` but the stored elements are plain `T`
```
lazarusA: the comment looks unnecessary, good to know this was the issue, but the code line below looks explicit enough.

meggart (approved):

lib/ZarrCore/src/pipeline.jl
asinghvi17:
```suggestion
```

PR #334, opened 2026-08-26
Fix complex v3 data_type names (complex64/complex128)

asinghvi17:
**Broken:** `ZarrCore/src/metadata3.jl:13` — the generic `typestr3(t::Type) = lowercase(string(t))` wrote `data_type` as `"complexf32"`/`"complexf64"` for `ComplexF32`/`ComplexF64`. The spec names are `"complex64"`/`"complex128"`, which is all `typemap3` reads, so a complex v3 array could never be reopened (`KeyError: key "complexf32" not found`).

**Fix:** explicit `typestr3(::Type{ComplexF32/F64})` methods returning the spec names, plus `typemap3` aliases for the old names so already-written stores still open.

Bundled: `fill_value_encoding(::Complex)` in `ZarrCore/src/metadata.jl` now emits the spec's `[real, imag]` instead of the JSON struct lowering `{"re","im"}`. v3 creation always synthesizes a `zero(T)` fill value, so without this no complex v3 array could be reopened and the `data_type` fix could not be tested end-to-end.

**Tests:** `Zarr/test/v3_codecs.jl` — `"complex data_type names (spec-compliant)"` (write, check `zarr.json`, round-trip via `zopen`) and `"complex data_type legacy-name read compat"` (hand-built `zarr.json` with the old names opens with the right eltype).

Validation after rebasing: `Zarr/test/v3_codecs.jl` passed all 211 checks on Julia 1.13.

Stacked on #339 (`top-level-package-workspace`).

https://claude.ai/code/session_011Qme1Mr9UxwceM15pWBG2k

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   `DateTime64`, `PermanentZarrCache`, `BloscCodec`, and `GzipCodec` through `Zarr`.
 - Declare an explicit public API [#317](https://github.com/JuliaIO/Zarr.jl/pull/317). Every store, codec, filter and compressor type, and every documented extension point, is now `public`; the set of exported names is unchanged. Internals (`Metadata`, `ZarrFormat`, `is_zarray`, `is_zgroup`, `normalize_path`, `MaxLengthString`, ...) are no longer reachable as `Zarr.x` and must be accessed via `Zarr.ZarrCore.x`
 - Fixed `fill_as_missing=true` on zarr v3 arrays, which threw `cannot reinterpret UInt8 as Union{Missing,Float64}` when decoding an initialized chunk.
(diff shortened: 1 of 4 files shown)


asinghvi17:
This got deslopified

lib/ZarrCore/src/metadata.jl
```diff
@@ -266,6 +266,8 @@ end
 
 fill_value_encoding(v) = v
 fill_value_encoding(::Nothing)=nothing
+# spec: complex fill values are `[real, imag]`; without this the JSON lowering is `{"re":..,"im":..}`
+fill_value_encoding(v::Complex) = Any[real(v), imag(v)]
```
lazarusA: why `Any`? doing `[real(v), imag(v)]` should do, no?

asinghvi17: It really should and I thought we merged this already.  But perhaps it got lost in the move.
```suggestion
fill_value_encoding(v::Complex) = [real(v), imag(v)]
```

meggart (approved):

PR #336, opened 2026-09-09
use accumulate when decoding DeltaFilter to match encode overflow

### LINKED ISSUE #335: Delta filter does not round trip for int types with overflow
```julia
julia> filter = Zarr.DeltaFilter{Int8}()
Zarr.DeltaFilter{Int8, Int8}()

julia> enc = Zarr.zencode(Int8[-128, 127, -128], filter)
3-element Vector{Int8}:
 -128
   -1
    1

julia> Zarr.zdecode(enc, filter)
ERROR: InexactError: trunc(Int8, -129)
Stacktrace:
  [1] throw_inexacterror(func::Symbol, to::Type, val::Int64)
    @ Core ./boot.jl:815
  [2] checked_trunc_sint
    @ ./boot.jl:829 [inlined]
  [3] toInt8
    @ ./boot.jl:844 [inlined]
  [4] Int8
    @ ./boot.jl:954 [inlined]
  [5] _broadcast_getindex_evalf
    @ ./broadcast.jl:699 [inlined]
  [6] _broadcast_getindex
    @ ./broadcast.jl:672 [inlined]
  [7] _getindex
    @ ./broadcast.jl:620 [inlined]
  [8] getindex
    @ ./broadcast.jl:616 [inlined]
  [9] macro expansion
    @ ./broadcast.jl:995 [inlined]
 [10] macro expansion
    @ ./simdloop.jl:77 [inlined]
 [11] copyto!
    @ ./broadcast.jl:994 [inlined]
 [12] copyto!
    @ ./broadcast.jl:947 [inlined]
 [13] copy
    @ ./broadcast.jl:919 [inlined]
 [14] materialize
    @ ./broadcast.jl:894 [inlined]
 [15] zdecode(data::Vector{Int8}, filter::Zarr.DeltaFilter{Int8, Int8})
    @ Zarr ~/.julia/packages/Zarr/BbOdA/src/Filters/delta.jl:33
 [16] top-level scope
    @ REPL[136]:1
```

I think this is related to what the docs for cumsum say:

```
help?> cumsum
search: cumsum cumsum! sum

[...]

  │ Note
  │
  │  The return array's eltype is Int for signed integers of less than system word size and UInt for unsigned
  │  integers of less than system word size. To preserve eltype of arrays with small signed or unsigned
  │  integer accumulate(+, A) should be used.
  │
  │  julia> cumsum(Int8[100, 28])
  │  2-element Vector{Int64}:
  │   100
  │   128
  │  
  │  julia> accumulate(+,Int8[100, 28])
  │  2-element Vector{Int8}:
  │    100
  │   -128
  │
  │  In the former case, the integers are widened to system word size and therefore the result is Int64[100,
  │  128]. In the latter case, no such widening happens and integer overflow results in Int8[100, -128].
```

Because `diff` _does_ overflow, we also I think want the decode to overflow as well; otherwise the transformation isn't reversible.

I'm pretty sure that the python numcodecs implementation (which just uses `np.cumsum`) also has the reversible overflow behavior, since I caught this in tests migrating from python Zarr to Zarr.jl, and my python decoder does decode the julia-serialized array.

kleinschmidt:
Fixes #335

### DIFF
--- a/test/Filters.jl
+++ b/test/Filters.jl
@@ -134,7 +134,8 @@ end
         Int32.(collect(0:999)),  # np.arange(1000, dtype='<i4')
         Float32.(reshape(LinRange(1000, 1001, 1000), (100, 10))),  # np.linspace(1000, 1001, 1000, dtype='<f4').reshape(100, 10)
         Float64.(reshape(randn(1000) .* 1 .+ 1000, (10, 10, 10))),  # np.random.normal(loc=1000, scale=1, size=(10, 10, 10)).astype('<f8')
-        permutedims(reshape(rand(UInt16, 1000) .% 200, (100, 10)))  # np.random.randint(0, 200, size=1000, dtype='u2').astype('<u2').reshape(100, 10, order='F')
+        permutedims(reshape(rand(UInt16, 1000) .% 200, (100, 10))),  # np.random.randint(0, 200, size=1000, dtype='u2').astype('<u2').reshape(100, 10, order='F')
+        Int8[-128, 127]
     ]
 
     for array in arrays
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 - Move code to ZarrCore.jl with low dependencies
 - Declare an explicit public API [#317](https://github.com/JuliaIO/Zarr.jl/pull/317). Every store, codec, filter and compressor type, and every documented extension point, is now `public`; the set of exported names is unchanged. Internals (`Metadata`, `ZarrFormat`, `is_zarray`, `is_zgroup`, `normalize_path`, `MaxLengthString`, ...) are no longer reachable as `Zarr.x` and must be accessed via `Zarr.ZarrCore.x`
+- Use overflowing `accumulate(+, ...)` instead of `cumsum` in `DeltaFilter` decode, to match overflowing behavior of `diff`.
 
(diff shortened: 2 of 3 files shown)


test/Filters.jl
```diff
@@ -134,7 +134,8 @@ end
         Int32.(collect(0:999)),  # np.arange(1000, dtype='<i4')
         Float32.(reshape(LinRange(1000, 1001, 1000), (100, 10))),  # np.linspace(1000, 1001, 1000, dtype='<f4').reshape(100, 10)
         Float64.(reshape(randn(1000) .* 1 .+ 1000, (10, 10, 10))),  # np.random.normal(loc=1000, scale=1, size=(10, 10, 10)).astype('<f8')
-        permutedims(reshape(rand(UInt16, 1000) .% 200, (100, 10)))  # np.random.randint(0, 200, size=1000, dtype='u2').astype('<u2').reshape(100, 10, order='F')
```
lazarusA, line 137: shouldn't this old test also pass?

kleinschmidt: yes it should! I just added a comma to continue the list.

kleinschmidt: ah I also mixed up the order of arguments in `accumulate!` 🫠

lazarusA (approved):

asinghvi17 (approved):

asinghvi17:
Thanks!  will merge after the subpackage split.

PR #344, opened 2026-09-20
Fix path-traversal check and DirectoryStore serving in zarr_req_handler

asinghvi17:
Two pre-existing bugs in `ZarrHTTP.zarr_req_handler` (both already in Zarr v0.10.2), fixed together because the second one was masking the first.

| Bug | Cause | Fix |
| --- | --- | --- |
| `/../secret.txt` is not rejected | `contains("..", k)` has its arguments swapped, so it tests whether `".."` contains the key | Validate the key with `invalid_key_reason` (below) and answer `400` instead of returning `nothing` from the handler |
| Serving a `DirectoryStore` at the root path fails on every request, even `/.zarray`, with `getindex not implemented for store DirectoryStore` | `lstrip` returns a `SubString`, and `DirectoryStore` only defines `getindex` for `String` (as do `ConsolidatedStore`, `S3Store` and `GCStore`) | Convert the key to a `String` in the handler |

## Key validation

`invalid_key_reason(store, path, key)` returns why a key is refused, or `nothing`. It runs two checks:

1. **All stores:** reject any key with a `..` path segment (split on `/` and `\`). The check matches whole segments, so a key such as `a1/0..0` is still served.
2. **`DirectoryStore`:** resolve the file the store would read (`abspath` of `folder/path/key`) and reject it unless it lies inside `folder/path`. This does not depend on spotting `..` in the key, and it also covers keys that `joinpath` treats as absolute, such as `C:/secret.txt` on Windows. `ConsolidatedStore` and `CachingStore` defer to the stores they wrap.

The `400` body names the request target and the reason, for example:

```
Error: Invalid request target "/../secret.txt": keys must not contain ".." path segments, since they could address data outside the served path
```

## Tests

New `"Serving a DirectoryStore"` testset in `Zarr/test/storage.jl`:

- round-trips a group and array through `HTTP.serve!` on a `DirectoryStore`
- calls the handler directly (HTTP clients normalize `..` away) with `/../secret.txt`, `/a1/../../secret.txt` and `/..\secret.txt`, and checks each gets `400` with the reason in the body
- checks the resolved-path rule on a `DirectoryStore` and through a `ConsolidatedStore`, using an absolute key that has no `..` segment

https://claude.ai/code/session_016rS9chzPZsektCcysh42ur


lazarusA:
I don't think any of the reasons given above for the bugs are true. That example might not work, but not for the reasons given in the table.

meggart (approved):

